PR #12, opened 2025-09-11
fix: Split out and fix wildcard utils with improve tests

machshev:
There are a lot of functions in the `utils.py` module. This PR starts splitting these out into separate modules and expanding the testing. This makes it easier to find utility functions and highlights where the test coverage gaps are.

The existing `utils.py` is now `utils/__init__.py` to allow the utils to be split into modules. This means that users of the utility functions can still import them from the same module path:

```
import dvsim.utils
from dvsim.utils import ...
```

The wildcard substitution tests `utils_tests.py` have been moved to the `tests` dir and expanded. The wildcard functions have also been refactored resolving some issues found while improving the testing. Further improvements could be made.


src/dvsim/utils/wildcards.py
```diff
@@ -0,0 +1,375 @@
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Wildcard substitution in strings within data structures.
+
+DVSim configuration relies heavily on wildcard based substitutions and dynamic
```
rswarbrick, line 7: I think that "wildcard" is a bit of a term of art from Sri. Maybe this could be written as
```
DVSim configuration relies heavily on dynamic templating and substitutions based on "wildcards".
```
?

rswarbrick (approved):
This is really nice! It's exciting to see things getting more structured.

PR #29, opened 2025-09-25
Improve logging

machshev:
The purpose of this PR is twofold:

- [x] Move the logging setup out of the cli main - this will simplify the main function and help later refactoring.
- [x] Standardise logging over the codebase - there is a lot of default logger use which is bad.
 
Python programs and especially libraries should not use the default logger but create a logger instance. If you use the default logger, then setting the log level for example would set the log level for any 3rd party libraries used (polluting the logs and making them difficult to read). Typically a logger instance would would be crated at module level with something like `logger = logging.getLogger(__name__)`. For the moment though, lets start with something simpler and make it more complicated later if desired.

The `LOG015` lint rule checks for default logger use instances. In the DVSim codebase that accounts for `129` instances (out of `1128` lint fails). With this PR the total lint fails is reduced to a mere `968` ;)


machshev:
It looks like later versions of ruff have introduced some extra linting rules that now fail. I've added a commit that resolves the failed lints, although not directly related to this PR.

rswarbrick (commented):
These look like nice cleanups! It's lovely to see things getting more solid.

PR #38, opened 2025-10-02
[launcher] Fake launcher to produce random results

machshev:
Add a fake luncher to generate random results instead of launching the actual deployment job.

This fake launcher is one step further than the dry-run mode and allows us to run a full flow through to the end, exercising everything apart from executing the jobs themselves. Making it easier to find issues with with DVSim with a quick local run without having to run a full regression.

Another use for this is generating fake run data for developing report templates for example. As well as investigating the scheduler behaviour without having to spawn long running sub processes. Perhaps useful for testing the scheduler behaviours.


mkj121 (approved):

KinzaQamar (approved):
LGTM, thanks!

engdoreis (approved):
LGTM.

Issue #34, opened 2025-10-01
Config loading regression test


machshev:
We need to know if changes in the config loading causes regressions with OT config. One suggestion is to add a mechanism to dump the parsed config during a DVSim run, then compare the before and after to ensure no regressions have been introduced during a refactor.

This is not as simple as it sounds, due to the wildcard substitution mechanism within the HJSON. Which is resolved in a cascading manor throughout the DVSim Job preparation step. So for a more robust check we need an end to end test for the broadest coverage. Then perhaps refine this further with integration tests on specific behaviours.

## Proposal

Add JSON dump mechanism to the DVSim job scheduler - so we can save the config of each job about to be run. Shouldn't be more than couple of lines of additional code (turned on with an environment variable flag). This is the simplest place I can think of for dumping the output of the config loading stage. It means we then have plain text diff-able artefact representing the actual jobs DVSim plans to run which can be compared for A/B testing.

This means we can do a manual A/B test and know if the stand-alone DVSim changes anything meaningful in terms of what and how the simulations are run.

PR #39, opened 2025-10-03
feat: add deployment object dump debug feature

machshev:
Debug feature to dump a JSON representation of the `Deploy` objects, just before they are sent to the scheduler. This feature is hidden behind an environment variable to avoid cluttering the CLI interface. The full `Deploy` object is not JSON serialisable in it's current form, so we only serialise the attributes that the LocalLauncher uses.

To enable set `DVSIM_DEPLOY_DUMP=true`, run DVSim and there will be a file called `deploy-[timestamp].json` generated in the scratch directory.

This file can they be compared between runs to check for regressions in the flow leading up to job deployment. With `--fixed-seed=1 --branch=baseline`, the resulting json files should be identical. Setting the "branch" doesn't actually change the git branch, it just overrides what DVSim thinks is the branch as far as paths in the scratch directory are concerned. We need to either set this to something fixed so that the deployment objects contain the same paths, otherwise a diff will fail.

Generating a `diff` or hash of the two files and comparing shows if the job deployments would be identical or not.

Using this functionality I have confirmed, by backporting this commit to the first release tag, that none of the refactorings made so far in this repository have changed the deployment objects in such a way that the launched jobs are different in any way.

```
✦ ❯ sha512sum baseline.json deploy_25.10.03_11.35.33.json
af732c3011753cfc7dc40068e1ce9b6cf3624973ffbbd25f828e4d7727f27dd65b5ada19407500315ea6f279946222d185dc939fe4978b0b32b3e7e8b4f7f60c  baseline.json
af732c3011753cfc7dc40068e1ce9b6cf3624973ffbbd25f828e4d7727f27dd65b5ada19407500315ea6f279946222d185dc939fe4978b0b32b3e7e8b4f7f60c  deploy_25.10.03_11.35.33.json
```

The first JSON file was generated from backporting the this feature to the tagged commit. The second file was generated with this branch and includes all the tidyup work made so far to DVSim.

The DVSim command used is:

```
DVSIM_DEPLOY_DUMP=true dvsim \
    ~/base/opentitan/hw/top_earlgrey/dv/top_earlgrey_sim_cfgs.hjson \
    --proj-root ~/base/opentitan -i nightly --scratch ~/scratch \
    --branch baseline --fixed-seed 1 \
    --verbose=debug --max-parallel=50 \
    --cov
```

**Note:** *your hashes will be different from mine as the directory paths will be different. However if you run this against the same versions your hashes should be the same as each other.*


mkj121 (approved):
Pirmin is away until Oct 20 so I suggest you remove him as a reviewer

machshev:
Turns out that the coverage merge directories are provided in a semi random order. Which means that if you run DVSim several times then the hash can be different due to that.

I've updated the PR to sort the deployment object model dump JSON objects by the job full name to make sure the jobs themselves are in the same order each time. However small variations may be present anyway due to issues like the above.

We could make sure these directory paths are sorted to make them more deterministic. But that's for another PR. For the moment this mechanism seems good enough.

hcallahan-lowrisc:
> Thanks @hcallahan-lowrisc 👍
> 
> I'm wondering about setting up sphinx to generate API docs? Then we could add this kind of thing in a debugging/developer page. Some architecture diagrams wouldn't go amiss either.

Yes, yes, yes. All the yesses. API docs would be amazing.

> Vaguely recollect seeing something like this in the main opentitan docs. However if we have to maintain it there as well as here it's going to diverge. Might be worth migrating at least the bits that make sense?

Good point, I think there is still some dvsim-specific markdown docs in the OT repo. We should definitely migrate it.

machshev:
This resolves #34

PR #45, opened 2025-10-29
Refactor scheduler and add tests for CompileSim deployment

machshev:
These are partial steps towards the goal of converting all the deployment objects to Pydantic models - which would provide the first complete and clear statically defined interface within the flow. From there it's a lot easier to work backwards and refactor the rest of the flow.

Main steps in this PR:

- [x] Make all the deployment objects attributes deterministic - prior to this change, the `cov_db_dirs` field order changed semi randomly. Meaning it takes manual review to determine if the `json` file generated with `DVSIM_DEPLOY_DUMP=true` changed due to reordering or an actual regression. Now a sha1 change is enough to warn of a breaking change.
- [x] Renamed `model_dump` to `dump` as the pydantic version contains intermediary fields and working data. While the dump script only exports the fields read by the scheduler and so affect the jobs that run. Once all deployment objects have been refactored into Pydantic then we could probably go back to `model_dump` or potentially have a single model of deployable Job to feed to the scheduler and vary only in the factory functions to generate them. Fundamentally the output is to the scheduler is little more than a "command to run", and environment variables to set in the subshell.
- [x] Add `WorkspaceCfg` to package up the values that were otherwise being read from the `Deploy` object. This means the `Deploy` object doesn't need to be passed around so much. This isn't necessarily the ideal end point and may disappear in later refactoring. However for the moment it's a step towards clearer interfaces and separation between data and behaviour.
- [x] Initial CompileSim unit tests - capturing the construction of the key deployment outputs consumed by the scheduler. This is to provide a harness for later refactoring.
- [x] Various small non functional improvements in the LSF launcher

### DIFF
--- a/src/dvsim/launcher/base.py
+++ b/src/dvsim/launcher/base.py
@@ -163,8 +163,8 @@ class Launcher(ABC):
         if not venv_path:
             venv_path = os.environ.get(common_venv)
 
-        if not Launcher.pyvenv:
-            Launcher.pyvenv = os.environ.get(f"{project.upper()}_PYVENV")
+        if venv_path:
+            Launcher.pyvenv = Path(venv_path)
 
     @staticmethod
     @abstractmethod
--- a/src/dvsim/launcher/lsf.py
+++ b/src/dvsim/launcher/lsf.py
@@ -52,22 +52,25 @@ class LsfLauncher(Launcher):
         """
         # Since we dispatch to remote machines, a project specific python
         # virtualenv is exists, needs to be activated when launching the job.
-        Launcher.set_pyvenv(project)
-        if Launcher.pyvenv is None:
+        Launcher.set_pyvenv(cfg.project)
+
+        pyvenv = Launcher.pyvenv
+        if pyvenv is None:
             return
 
         # If it is already a dir, then nothing to be done.
-        if os.path.isdir(Launcher.pyvenv):  # noqa: PTH112
+        if pyvenv.is_dir():
             return
 
         # If not, then it needs to be a valid tarball. Extract it in the
         # scratch area if it does not exist.
-        stem = Path(Launcher.pyvenv).stem
+        stem = pyvenv.stem
         if stem.endswith("tar"):
             stem = stem[:-4]
-        path = Path(args.scratch_root, stem)
+
+        path = cfg.scratch_root / stem
         if not path.is_dir():
-            log.info("[prepare_workspace]: [pyvenv]: Extracting %s", Launcher.pyvenv)
+            log.info("[prepare_workspace]: [pyvenv]: Extracting %s", pyvenv)
             with tarfile.open(Launcher.pyvenv, mode="r") as tar:
                 tar.extractall(cfg.scratch_root)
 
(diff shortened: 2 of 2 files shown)


src/dvsim/launcher/base.py
```diff
@@ -90,6 +90,48 @@ class Launcher(ABC):
         context=[],
     )
 
+    def __init__(self, deploy: "Deploy") -> None:
```
rswarbrick: Do we need the quotes here? I'd have thought that it only matters when we're type checking, in which case it's in scope?

src/dvsim/sim_utils.py
```diff
@@ -102,17 +112,23 @@ def vcs_cov_summary_table(buf):
...
-    `log_text` is the job's log file contents as a list of lines.
-    `tool` is the EDA tool used to run the job.
-    Returns the runtime, units as a tuple.
-    Raises NotImplementedError exception if the EDA tool is not supported.
+    Args:
+        log_text: is the job's log file contents as a list of lines.
+        tool: is the EDA tool used to run the job.
+
+    Returns:
+        the runtime, units as a tuple.
```
rswarbrick: This tripped me up slightly. How about: "a tuple of (runtime, units)." ?

I originally understood it to be talking about "units as a tuple" which isn't quite right :-)

src/dvsim/launcher/local.py
```diff
@@ -181,3 +182,24 @@ def _close_job_log_file(self) -> None:
         """Close the file descriptors associated with the process."""
         if self._log_file:
             self._log_file.close()
+
+    @staticmethod
```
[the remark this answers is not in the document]
machshev: Probably not, at least in it's current implementation. They are called by the launcher base class as part of the launcher interface. It gives the launcher the opportunity to prepare the workspace if required, otherwise it's just a no-op.

If we raised in the method it would break the launcher. Unless we added try/except to the call... which wouldn't be a good idea.

I'm not keen on this mechanism, I think from memory it's only used by one Launcher type. So there may be a better way of implementing it while keeping the interface smaller for the general case.

Either way, refactoring the launchers is quite risky right now. We don't have a way of testing all of them, and there are no unit-tests as some form of safety net. I'm hoping to add those at some point, at which point we can be more adventurous in our refactoring. At the moment I'm just trying to make interfaces more obvious... then we can have discussions like this on how much sense they make ;)

src/dvsim/sim_utils.py
machshev: Yea, I'm finding a lot of the docstrings here confusing. Often having to resort to reading the code.

src/dvsim/launcher/base.py
rswarbrick: I agree with both of those things. My question was why the quotes are needed in this particular case.

But I think I've realised: even if the type annotation doesn't do anything at runtime, all of its symbols still need to be in scope.

rswarbrick (commented):
Whoah! Please could you apply the changes requested by the review to the commits they were talking about? Don't forget that we don't use squash-merging.

Also, it would be really helpful if you could respond to a review comment (or add a thumbs up, or not bother saying anything specifically on the comment) BUT NOT MARK IT RESOLVED. Otherwise the poor reviewer has to manually wade through all the resolved comments and look to figure out whether the requested change had actually been made. It's rather inefficient.

With a big long list of comments on a PR of mine, my approach is usually:
- Hack away in Emacs, using force push locally and interactive rebase to address the changes the reviewer suggested.
- If they are talking nonsense, reply to the review comment in question. ("No! That's a terrible idea!" :-D)
- (Ideally) run the tests again locally to check that I didn't just break compliation with a typo.
- Force-push the new version.
- Respond to the reviewer with a "I think your review is beautiful" (or something more honest)
- Re-request review.

If the changes are small, the reviewer can then hit the "Compare" button next to my force-push to see what I changed. To make *that* work, it's helpful not to rebase onto master (because then the comparison picks up all of those changes). If I really want to rebase onto master, I'm sometimes really enthusiastic and do it in two stages:
- Rebase onto master.
- Force-push and leave a comment on the PR saying "No functional change: I've just rebased."
- Make whatever tweak I was trying to make.
- Force-push again, and leave a comment on the PR saying "I did something this time"

(Of course, I normally forget to do it in that order. You can totally do it the other way around too, and make the tweak first :-D)

Issue #53, opened 2025-11-12
Summary report test filtering


machshev:
https://nightly.reports.lowrisc.org/opentitan_nightly_darjeeling/2025_11_11/hw/top_darjeeling/dv/summary/report.html

The underlying test reports seem to contain all tests, rather than filtering to the IP under test. Which means the summary report summarises the full set of tests for each IP.

PR #54, opened 2025-11-12
[reporting] fix various bugs

machshev:
- [x] **Report item filtering:** the `CompletedJobStatus` objects were no longer being filtered by project. So each report page had the entire set of tests.
- [x] Max job runtime and Simulated time - units added and precision fixed to 3 decimal places.

Fixes: #53


hcallahan-lowrisc (approved):
LGTM, thanks for the fixes!

PR #63, opened 2025-11-19
Report fixes

machshev:
Some bug introduced with the new report template.

- [x] Fix JSON link on summary page
- [x] Variant missing

Fixes: #65

### DIFF
--- a/src/dvsim/templates/reports/summary_report.html
+++ b/src/dvsim/templates/reports/summary_report.html
@@ -106,7 +106,7 @@
                         <tr>
                             <td>
                                 <a href="{{ block_name }}.html">
-                                    {{ block_name }}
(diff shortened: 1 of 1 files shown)


machshev:
Thanks @etterli and @martin-velay!
It's a step in the right direction hopefully. I'm slowly moving the data into pydantic data classes and hopefully grouping them more logically. There is currently quite a bit of duplication in the data classes, so it would be nice to consolidate a bit more in the future.

Additionally some of the data should be calculated earlier in the flow and then referenced. But due to the way the wildcards are calculated that isn't easy at the moment. Hopefully when the earlier stages of the flow can be refactored this will be possible.

The report generation bug appears to be resolved with these changes when I test it locally.

PR #74, opened 2025-12-01
fix: Don't use python311 datetime alias

hcallahan-lowrisc:
Does not work for python310, which we need to support.


machshev:
I'm surprised the unit tests don't pick this up. They should run with 3.10 https://github.com/lowRISC/dvsim/blob/9f09c8f9359e8583cb95f2c6c8912bdef34f0c7a/.github/workflows/ci.yml#L87

PR #81, opened 2025-12-15
HTMX wrapper for the reports

machshev:
We want to be able to generate reports that contain the same data but switch the CSS theme depending on the website they are embedded within.

## Approach

[HTMX](https://htmx.org/) is a nice way of doing this without having to write an maintain JS. It allows declarative HTML to make requests to a server replace the contents of an element with the returned content.

Typically to implement something like this, we would have some JS make a request to a backend server which returns some data and then serialise the data and render it client side. With the HTMX approach, events can be defined to trigger a request to a server which returns pre-rendered HTML which is substituted in place. This is an interesting approach as it keeps the state server side, avoiding issues related to synchronising state. In practice this approach can be faster as well, given that server side the operations required to calculate the data are often more resource hungry than the rendering... an there isn't necessarily that much difference between the effort to render HTML fragments verses JSON.

**_In our case there will be no back-end rendering as we are just presenting static HTML. So the main use of HTMX is it's ability to fetch HTML fragments and insert it into the DOM._**

## Implementation

The intention is to create an HTMX wrapper page that is the actual `index.html` and then use it to embed the content of the individual report pages. The CSS can go in the wrapper, which means other wrappers can embed the same report with a different CSS script.

### DIFF
--- a/src/dvsim/templates/reports/wrapper.html
+++ b/src/dvsim/templates/reports/wrapper.html
@@ -4,34 +4,32 @@
 # SPDX-License-Identifier: Apache-2.0
 
 -->
-<!doctype html>
+<!DOCTYPE html>
 <html lang="en">
-<head>
-    <meta charset="utf-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1">
-    <title>Simulation Results</title>
-    <script src="js/htmx.min.js"></script>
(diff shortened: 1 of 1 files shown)


machshev:
Minimal visual difference:
image

The only known visual difference is the URL links being replaced with bootstrap buttons. Providing a nicer presentation by default and still giving feedback on mouse over.

PR #91, opened 2026-01-30
Enlighten status printer fixes

AlexJones0:
A couple of small bug fixes / cleanups to the `EnlightenStatusPrinter` - see the commit messages for more details. There is definitely potential for refactoring the help message logic but the change is kept intentionally small for now to focus on the bug fix, and because larger-scale refactors/rewrites may come later.

### DIFF
--- a/src/dvsim/utils/status_printer.py
+++ b/src/dvsim/utils/status_printer.py
@@ -202,6 +202,7 @@ class EnlightenStatusPrinter(TtyStatusPrinter):
             self.status_header.close()
         for target in self.status_target:
             self.status_target[target].close()
+        self.manager.stop()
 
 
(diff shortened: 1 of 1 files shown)


src/dvsim/utils/status_printer.py
```diff
@@ -23,7 +26,7 @@ class StatusPrinter:
     def __init__(self) -> None:
         """Initialise."""
 
-    def print_header(self, msg: str) -> None:
+    def print_header(self, msg: str = DEFAULT_HEADER) -> None:
```
rswarbrick: I'm not a big fan of optional arguments. Maybe leave this compulsory and add a `print_default_header()` function?

machshev (approved):
Nice, thanks @AlexJones0

PR #93, opened 2026-02-09
Add many scheduler unit & integration tests

AlexJones0:
This PR introduces a large number of tests specifically targeting the scheduler. The primary goals are to:
1. Provide a better understanding of the existing features of the scheduler.
2. Show a clearer picture of what is and isn't working within the scheduler, so we know what may need to be fixed.
3. Enable refactoring and/or rewriting of the scheduler with reduced chances of introducing unintentional changes or regressions in functionality.

To those ends, this PR:
1. Creates mocks of the scheduled jobs / `Launcher` class to allow us to test the behaviour of the scheduler, by querying the mocked data.
2. Introduces additional `test` dev dependencies and project/CI config to allow better testing of the scheduler.
3. Adds a wide variety of tests covering: smoke tests, job launching / polling, job weighting / priority, parallelism, dependency resolution, error cases, edge cases, etc. as well as tests for the scheduler's signal handlers.

See the commit messages and comments for more information. The intention is to address issues in follow up PRs (either with bug fixes or complete refactors/rewrites), rather than overloading this PR with even more content.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -133,6 +133,13 @@ ignore = [
 addopts = "--cov=dvsim --cov-report term-missing"
 norecursedirs = ["*.egg", ".*", "_darcs", "build", "dist", "venv", "scratch", "doc"]
 
+[tool.coverage.run]
+parallel = true
+concurrency = ["multiprocessing"]
+
+[tool.coverage.report]
+show_missing = true
+
 [tool.semantic_release]
 commit_parser = "conventional"
 version_toml = ["pyproject.toml:project.version"]


tests/test_scheduler.py
```diff
@@ -193,3 +198,230 @@ class Fxt:
...
+
+def job_spec_factory(
+    *,
+    paths: JobSpecPaths | None = None,
+    deps: list[str] | None = None,
+    **overrides: object,
+) -> JobSpec:
+    """Create a JobSpec from a set of default values, for use in testing."""
+    output_dir = paths.output if paths else Path("/tmp/scheduler_output")
+    log_path = paths.log if paths else Path("/tmp/scheduler.log")
```
machshev: Again I'd avoid global `/tmp` by making paths compulsory. It makes it easier to reason about if the directory is always the same.

tests/test_scheduler.py
```diff
@@ -193,3 +198,230 @@ class Fxt:
...
+    links = paths.statuses if paths else {}
+    spec = {
+        "name": "test_job",
+        "job_type": "mock_type",
+        "target": "mock_target",
+        "seed": None,
+        "block": ip_meta_factory(),
+        "tool": tool_meta_factory(),
+        "workspace_cfg": workspace_config_factory(),
```
machshev: This looks like the workspace config paths can be different from the paths in `paths`. If these are the same then maybe create one from the other?

tests/test_scheduler.py
```diff
@@ -425,3 +453,536 @@ def test_many_jobs_run(fxt: Fxt) -> None:
...
+    @pytest.mark.timeout(DEFAULT_TIMEOUT)
+    def test_no_over_poll(fxt: Fxt) -> None:
+        """Test that the schedule stops polling when it sees `PASSED`, and does not over-poll."""
+        jobs = make_many_jobs(fxt.tmp_path, 10)
+        polls = [(i + 1) * 10 for i in range(10)]
+        for i in range(10):
+            mock = [(0, JobStatus.DISPATCHED), (polls[i], JobStatus.PASSED)]
+            fxt.mock_ctx.set_config(jobs[i], MockJob(status_thresholds=mock))
+        scheduler = scheduler_factory(jobs, fxt.mock_launcher)
+        result = scheduler.run()
```
machshev: Maybe
```
result = scheduler_factory(jobs, fxt.mock_launcher).run()
```
The scheduler is not used after this so may as well chain and reduce the intermediate variables.
For tests I'd aim for this more than the main code, as it just makes it easier to read the tests and understand what's being tested.

tests/test_scheduler.py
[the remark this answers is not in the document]
AlexJones0: I've removed `WorkspacePaths` and the function for creating it, but I'm leaving `timestamp` as a string with a default value here for now to minimize the amount of refactoring changes in this PR.

PR #96, opened 2026-02-11
Add DVSim version to generated reports

AlexJones0:
One thing I noticed when looking at the output reports is that DVSim doesn't mention what version of DVSim the report was generated with, despite already providing the commit of the top being tested (for sim flows at least). Since DVSim is undergoing active changes, it seems prescient to record this in the generated reports for the sake of reproducibility.

This PR adds the version that was used to run DVSim to the generated block and summary reports, with a link to the tag of the relevant Github release for that version.

### DIFF
--- a/src/dvsim/sim/data.py
+++ b/src/dvsim/sim/data.py
@@ -233,6 +233,9 @@ class SimResultsSummary(BaseModel):
     top: IPMeta
     """Meta data for the top level config."""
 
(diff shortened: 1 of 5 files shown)


hcallahan-lowrisc (approved):

machshev (approved):
Nice, Thanks @AlexJones0!

PR #114, opened 2026-03-11
Remove duplicate data from summary

machshev:
Prepare for adding support for a dashboard.

- [x] Remove the old FlowResults class - residual from previous refactoring
- [x] Remove duplicate data from the simulation results summary
- [x] Fix a bug in the markdown report paths


AlexJones0 (commented):
Thanks for the work on this @machshev, it definitely seems like an improvement.

I'm happy for this to go in when the Python versioning issue is addressed.

PR #115, opened 2026-03-11
Dashboard

machshev:
Add support for generating a dashboard.

### DIFF
--- a/src/dvsim/sim/dashboard.py
+++ b/src/dvsim/sim/dashboard.py
@@ -1,0 +1,61 @@
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Generate dashboard.
+
+The dashboard is a cut down version of the full report where a simpler summary
+is required than the full report simulation summary. This is intended to be used
+on a separate website and links back to the detailed report if required.
+
+This is intended to generate a dashboard that could be used on the OpenTitan
+and automatically i.e. https://opentitan.org/dashboard/index.html
+"""
+
+from pathlib import Path
+
+from dvsim.logging import log
+from dvsim.report.artifacts import render_static_content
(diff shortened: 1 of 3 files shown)


src/dvsim/templates/dashboard/index.html
```diff
@@ -0,0 +1,110 @@
...
+                                <th class="bg-secondary text-white">Assertion</th>
+
+                            </tr>
+                        </thead>
+                        <tbody>
+                            {% for block_name in summary.flow_results.keys()|sort %}
+                                {% set flow = summary.flow_results[block_name] %}
+                                <tr>
+                                    <td class="p-0">
+                                        <a href="{{ base_url }}{{ block_name }}.html"
```
AlexJones0: If it isn't too hard, can you either:
1. Omit the button / hyperlink if no base URL is given, or
2. Get the base URL from the `index.json` path parent if it is not provided.

Depending on whatever the intended behaviour is, because at the moment if I run this locally without specifying `--base-url` I just get broken links. I suspect the intended behaviour is (1), so that if we generate a dashboard without giving it a link to find block reports (e.g. on a hosted site), it just doesn't link anywhere.

I think it would be better to have `base_url` be `str | None` (default `None`), and then only link in the template if `base_url is not none`.

src/dvsim/sim/dashboard.py
```diff
@@ -0,0 +1,53 @@
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Generate reports."""
```
AlexJones0: Also, I think this module-level comment needs to be clearer here (it seems like a copy-paste?)

Maybe something like "Generate simplified result dashboards" or the like?

AlexJones0 (commented):
I appreciate that all the changes may not yet have been made, but I've left a couple more nits for the latest pushes (sorry!)

PR #136, opened 2026-04-01
Convert scheduler tests to use the new scheduler

AlexJones0:
This PR is the fourteenth of a series of PRs to rewrite DVSim's core scheduling functionality (Scheduler, status display, launchers / runtime backends) to use an async design, with key goals of long term maintainability and extensibility.

This PR converts the scheduler tests to use the new async scheduler so that we can test the new design. As CI shows, the scheduler now passes _all_ of the existing scheduler tests (plus, a couple of new tests, added in this PR). The intention is not to merge this PR yet - it should probably wait until all the other scheduler integration is completed and merged, and should only be merged right before making the switch between the old and new scheduler. Instead, this PR is intended to show the correctness and functionality of the new scheduler via the existing tests.


AlexJones0:
The last few force pushes have all been rebases - nothing else has changed.

PR #142, opened 2026-04-02
[src] Integrate DVPlan tool commands

martin-velay:
- Integrate DVPlan subcommands. These commands will be run only if:
    - DVPlan tool is available
    - Coverage is enabled
    - A "".vplan" file is found in the cuurent sim_cfg.hjson file
- When the condition are met, at the end the regression/simulation, the DVPlan tool will:
    - Check the coverage of the specification
    - Generate a derived vPlan in the simulation directory
    - Perform back-annotation of the vPlan with the coverage data
    - Compute the coverage of the vPlan
    - Generate a report based on obtained annotated vPlan

### DIFF
--- a/src/dvsim/job/deploy.py
+++ b/src/dvsim/job/deploy.py
@@ -30,9 +30,9 @@ if TYPE_CHECKING:
     from dvsim.modes import BuildMode
     from dvsim.sim.flow import SimCfg
 
-
 __all__ = (
     "CompileSim",
+    "CovVPlan",
     "Deploy",
 )
 
@@ -1028,3 +1028,80 @@ class CovAnalyze(Deploy):
         self.qual_name = self.target
         self.full_name = f"{self.sim_cfg.name}{self._variant_suffix}:{self.qual_name}"
         self.input_dirs += [self.cov_merge_db_dir]
+
+
+class CovVPlan(Deploy):
+    """Abstraction for generating a Verification Plan (vPlan) report using DVPlan."""
+
+    target = "cov_vplan"
+    weight = 10
+
+    def __init__(self, cov_report_job, sim_cfg) -> None:
+        self.report_job = cov_report_job
+        super().__init__(sim_cfg)
+        self.dependencies.append(cov_report_job)
+
+    def _define_attrs(self) -> None:
+        super()._define_attrs()
+        self.mandatory_cmd_attrs.update(
+            {
+                "proj_root": False,
+                "vplan": False,
+            }
+        )
+        self.mandatory_misc_attrs.update(
+            {
+                "dut_instance": False,
+            }
+        )
+
+    def _set_attrs(self) -> None:
+        self.cov_vplan_dir = f"{self.sim_cfg.scratch_path}/{self.target}"
+
+        super()._set_attrs()
+        self.qual_name = self.target
+        self.full_name = f"{self.sim_cfg.name}{self._variant_suffix}:{self.qual_name}"
+
+        self.prepare_opts = self.sim_cfg.cov_vplan_prepare_opts
+        self.process_opts = self.sim_cfg.cov_vplan_process_opts
+
+        # Calculate IP root.
+        vplan_path = Path(self.vplan)
+        self.ip_root = str(vplan_path.parent.parent)
+
+        # Use fixed output filenames so the report location is always predictable.
+        self.annotated_hjson = f"{self.odir}/vplan_annotated.hjson"
+        self.gen_html = f"{self.odir}/vplan_annotated.html"
+        self.output_dirs = [self.odir]
+
+    def _construct_cmd(self) -> str:
+        """Construct the pure bash shell command, bypassing the base Makefile assumption."""
+        import shlex
+        import shutil
+
+        if shutil.which("dvplan") is None:
+            fallback = (
+                "echo 'WARNING: dvplan tool not installed in PATH. Skipping vPlan generation.'"
+            )
+            return f"/usr/bin/env bash -c {shlex.quote(fallback)}"
+
+        def format_opts(opts):
+            return " ".join(opts) if isinstance(opts, list) else str(opts)
+
+        prepare_opts_str = format_opts(self.prepare_opts)
+        process_opts_str = format_opts(self.process_opts)
+
+        prepare_cmd = f"dvplan prepare_vplan {prepare_opts_str} {self.ip_root} {self.vplan} {self.annotated_hjson}"
+        prepare_cmd = " ".join(prepare_cmd.split())
+
+        vendor_tool = f"{self.sim_cfg.tool}_report"
+        report_path = self.report_job.cov_report_dir
+
+        process_cmd = (
+            f"dvplan process_results {process_opts_str} --coverage {vendor_tool} {report_path} "
+            f"-R {self.gen_html} -s {self.sim_cfg.name} {self.dut_instance} {self.annotated_hjson}"
+        )
+        process_cmd = " ".join(process_cmd.split())
+
+        full_command = f"set -e; mkdir -p {self.odir}; {prepare_cmd} && {process_cmd}"
+        return f"/usr/bin/env bash -c {shlex.quote(full_command)}"
--- a/src/dvsim/job/deploy.py
+++ b/src/dvsim/job/deploy.py
@@ -1038,6 +1038,8 @@ class CovVPlan(Deploy):
 
     def __init__(self, cov_report_job, sim_cfg) -> None:
         self.report_job = cov_report_job
+        # Populated by post_finish() once the job completes successfully.
+        self.vplan_coverage: float | None = None
         super().__init__(sim_cfg)
         self.dependencies.append(cov_report_job)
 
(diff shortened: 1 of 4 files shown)


machshev:
Looks like there is a Ruff lint issue?

martin-velay:
Hi, I have rebased on the master. Can you review my PR?

src/dvsim/job/deploy.py
```diff
@@ -1028,3 +1028,87 @@ def _set_attrs(self) -> None:
...
+    def _construct_cmd(self) -> str:
+        """Construct the pure bash shell command, bypassing the base Makefile assumption."""
+        import shlex
+        import shutil
+
+        if shutil.which("dvplan") is None:
+            fallback = (
+                "echo 'WARNING: dvplan tool not installed in PATH. Skipping vPlan generation.'"
+            )
+            return f"/bin/bash -c {shlex.quote(fallback)}"
```
machshev: `/bin/env bash` is better for distributions such as NixOS.

src/dvsim/job/deploy.py
```diff
@@ -1028,3 +1028,87 @@ def _set_attrs(self) -> None:
...
+        )
+
+        process_cmd = (
+            f"dvplan process_results {process_opts_str} --coverage {vendor_tool} {report_path} "
+            f"-R {self.gen_html} -s {self.sim_cfg.name} {self.dut_instance} {self.annotated_hjson}"
+        )
+        process_cmd = " ".join(process_cmd.split())
+
+        full_command = f"set -e; mkdir -p {self.odir}; {prepare_cmd} && {process_cmd}"
+        return f"/bin/bash -c {shlex.quote(full_command)}"
```
machshev: Same here

src/dvsim/sim/tool/base.py
```diff
@@ -95,3 +95,27 @@ def set_additional_attrs(deploy: "Deploy") -> None:
...
+
+    @staticmethod
+    def get_dvplan_parser_name() -> str:
+        """Return the name of the DVPlan parser for this tool.
+
+        Returns:
+            The DVPlan coverage parser string.
+
+        """
+        return "synopsys_report"
```
machshev: It might be nice to remove this function and just use `p"{self.sim_cfg.tool}_report"` instead. Which would mean you would need to update vplan to take "vcs_report" and "xcelium_report" instead. You could have an internal mapping in vplan to allow backward compatibility.

However long term it's a lot nicer to avoid having to maintain this extra method in the interface.

machshev: This is a Protocol not an abstract base class... so shouldn't have an implementation.
`...` should be the function body.

src/dvsim/sim/tool/base.py
```diff
@@ -95,3 +95,27 @@ def set_additional_attrs(deploy: "Deploy") -> None:
...
+
+        Args:
+            cov_report_dir: Path to the coverage report directory.
+            cov_report_txt: Path to the text-based coverage report.
+
+        Returns:
+            Path to the target coverage report file or directory.
+
+        """
+        return cov_report_txt or cov_report_dir
```
machshev: This is a Protocol not an abstract base class... so shouldn't have an implementation.
`...` should be the function body.

src/dvsim/sim/tool/xcelium.py
```diff
@@ -170,3 +170,27 @@ def set_additional_attrs(deploy: "Deploy") -> None:
...
+
+        Args:
+            cov_report_dir: Path to the coverage report directory.
+            cov_report_txt: Path to the text-based coverage report.
+
+        Returns:
+            Path to the target coverage report file or directory.
+
+        """
+        return f"$(ls -t {cov_report_dir}/report_data/*.report | head -n 1)"
```
machshev: This logic really belongs in the VPlan tool in the specific tool based parser.
I'd standardise on passing the `cov_report_dir` to VPlan and then the directory structure is known within VPlan.

src/dvsim/sim/data.py
```diff
@@ -192,6 +192,8 @@ class SimFlowResults(BaseModel):
     """Coverage metrics."""
     cov_report_page: Path | None
     """Optional path linking to the generated coverage report dashboard page."""
+    vplan_report_page: Path | None
```
machshev, line 195: Does this need to be a variable? Is it ever going to be anything other than a constant?
Unless there is a very good reason I'd standardise on a fixed relative path from the reports directory i.e. `p"{report_dir}/vplan/index.html"`.

I'd like to do the same with `cov_report_page` as well at some point i.e. `p"{report_dir}/cov/index.html"`.

src/dvsim/sim/flow.py
```diff
@@ -157,6 +158,12 @@ def __init__(self, flow_cfg_file, hjson_data, args, mk_config) -> None:
         self.cov_report_dir = ""
         self.cov_report_page = ""
 
+        # Options for vPlan processing
+        self.cov_vplan_dir = ""
+        self.dut_instance = ""
```
machshev, line 164: What is `dut_instance`? There is already a number of IP names, `name`, `qual_name`, `full_name`, ...
Is this adding anything new?

src/dvsim/sim/flow.py
```diff
@@ -157,6 +158,12 @@ def __init__(self, flow_cfg_file, hjson_data, args, mk_config) -> None:
         self.cov_report_dir = ""
         self.cov_report_page = ""
 
+        # Options for vPlan processing
+        self.cov_vplan_dir = ""
```
machshev: Can this be a constant relative directory?

src/dvsim/sim/flow.py
```diff
@@ -835,6 +850,7 @@ def make_test_result(tr) -> TestResult | None:
             stages=stages,
             coverage=coverage_model,
             cov_report_page=cov_report_page,
+            vplan_report_page=vplan_report_page,
```
machshev, line 854: Again... can this be a constant relative directory?
Unless there is a really good reason I'd prefer to just standardise on a directory structure.

We had an issue recently where we updated the config to provide a slightly different directory structure and even though it was configurable that way there were assumptions in the code that meant it didn't work. There were strange errors that took a while to work out. So if we can standardise on paths it really helps reduce the complexity.

The scratch directory structure is internal working data, so as long as it's logically structured I don't think it matters the exact structure. There is no reason I'm aware of for allowing flexibility in the structure, it just adds unnecessary complexity. The report directory structure is also better to standardise, as it makes it easier to test.

src/dvsim/sim/flow.py
```diff
@@ -550,6 +557,10 @@ def _create_deploy_objects(self) -> None:
                 self.cov_report_deploy = CovReport(self.cov_merge_deploy, self)
                 self.deploy += [self.cov_merge_deploy, self.cov_report_deploy]
 
+                if hasattr(self, "vplan") and self.vplan:
```
machshev: What is the type of this attribute? Is it a `bool`?
I'd probably use `if getattr(self, "vplan", False):` to make it a little clearer what the type is.

machshev (commented):
Thanks @martin-velay! Generally looks good, just a few things to tidy up ideally. It would really help to clean up the code if we can get these changes in.
The existing implementation of some of these parts of the code are already a bit of a mess, so this is a good opportunity to start putting in some better examples.

src/dvsim/sim/tool/xcelium.py
martin-velay: Not a bad idea, thanks

src/dvsim/sim/flow.py
martin-velay: `dut_instance` is the hierarchy path down to the DUT, it already exists in the related HJSON config file and is required. I think it differs from the one you mention. Is my understanding correct?

src/dvsim/sim/data.py
martin-velay: I don't mind having a constant name. I have made the change

src/dvsim/sim/flow.py
machshev: Thanks for explaining

src/dvsim/job/deploy.py
```diff
@@ -1035,3 +1028,85 @@ def _set_attrs(self) -> None:
...
+        process_opts_str = format_opts(self.process_opts)
+
+        prepare_cmd = f"dvplan prepare_vplan {prepare_opts_str} {self.ip_root} {self.vplan} {self.annotated_hjson}"
+        prepare_cmd = " ".join(prepare_cmd.split())
+
+        vendor_tool = f"{self.sim_cfg.tool}_report"
+        cov_report_txt = getattr(self.report_job, "cov_report_txt", "")
+        report_path = plugin.get_dvplan_report_path(
+            cov_report_dir=self.report_job.cov_report_dir, cov_report_txt=cov_report_txt
+        )
```
machshev, line 1103:
```suggestion
        report_path = self.report_job.cov_report_dir
```
Don't need to call the function that always returns the same value back.

src/dvsim/sim/tool/base.py
```diff
@@ -100,3 +95,17 @@ def set_additional_attrs(deploy: "Deploy") -> None:
 
         """
         ...
+
+    @staticmethod
+    def get_dvplan_report_path(cov_report_dir: str, cov_report_txt: str) -> str:
```
machshev, line 100: Not required any more

src/dvsim/sim/data.py
machshev: Can you remove this variable if it's now a constant? Just move the constant to where this variable is used?

src/dvsim/sim/tool/vcs.py
```diff
@@ -150,3 +145,17 @@ def set_additional_attrs(deploy: "Deploy") -> None:
             deploy: the deploy object to mutate.
 
         """
+
+    @staticmethod
+    def get_dvplan_report_path(cov_report_dir: str, cov_report_txt: str) -> str:
```
machshev, line 150: Not required anymore

src/dvsim/sim/tool/xcelium.py
```diff
@@ -176,3 +170,17 @@ def set_additional_attrs(deploy: "Deploy") -> None:
             deploy: the deploy object to mutate.
 
         """
+
+    @staticmethod
+    def get_dvplan_report_path(cov_report_dir: str, cov_report_txt: str) -> str:
```
machshev, line 175: Not required any more

src/dvsim/sim/flow.py
machshev: This line should be removed

src/dvsim/sim/data.py
machshev: Maybe we leave this one in for the moment and resolve it later with a follow up PR that also removes the `cov_report_page` as well.

src/dvsim/sim/flow.py
machshev: For the moment this could stay and we tidy this and the cov up as a follow-up PR.

machshev (approved):
Thanks @martin-velay

PR #143, opened 2026-04-02
Add experimental Z01X support to sync with in-tree OpenTitan DVSim

AlexJones0:
Since DVSim was pulled out and separated from OpenTitan, one additional PR has been made that has modified the operation of DVSim: https://github.com/lowRISC/opentitan/pull/28782

This PR aims to replicate the changes from that PR, refactored to fit out-of-tree DVSim. Two of the commits specifically implement the same functionality as from that PR - the other commits are related fixes not in the original PR.
1. For the second commit, almost nothing is changed, except for fixes to keep the `Deploy` tests passing.
2. For the fourth commit, the modification to `RunTest` is changed to now instead add `Z01X` as a separate simulation tool plugin. This way, we can inherit all of the relevant plugin functionality of VCS, whilst decoupling the Z01X-specific logic from the Deploy core itself. To do this, a generic mechanism is introduced to allow tools to mutate deploy objects after all initialization and expansion.

### DIFF
--- a/src/dvsim/sim/tool/base.py
+++ b/src/dvsim/sim/tool/base.py
@@ -44,10 +44,12 @@ class SimTool(Protocol):
 
         Args:
             log_text: is the job's log file contents as a list of lines.
-            tool: is the EDA tool used to run the job.
 
         Returns:
             a tuple of (runtime, units).
+
+        Raises:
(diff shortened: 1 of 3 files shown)


AlexJones0:
Closing this in favor of splitting into separate PRs.

PR #148, opened 2026-04-02
Allow the type of self.sim_cfg not to have commit/commit_short

rswarbrick:
**NOTE: This is a draft because it depends on #146 (the first commit in the PR)**

This wouldn't work at runtime, but postponing the check until then is a reasonable stop-gap that allows us to postpone working out which subclasses of FlowCfg define commit/commit_short until we have tidied up the types in more of those classes.

### DIFF
--- a/src/dvsim/job/deploy.py
+++ b/src/dvsim/job/deploy.py
@@ -113,6 +113,19 @@ class Deploy:
         # If the FlowCfg object in self.sim_cfg doesn't have a links dictionary, default to {}.
         links = getattr(self.sim_cfg, "links", {})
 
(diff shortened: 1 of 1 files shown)


AlexJones0 (changes requested):
Maybe I missed something, but I think this is now redundant after https://github.com/lowRISC/dvsim/pull/139 was merged earlier today, perhaps you need to pull in the latest `master`? That PR moved the `commit`/`commit_short` attributes to be defined on the base `FlowCfg`, so now I would expect it should always exist and the typing shouldn't complain.

PR #156, opened 2026-04-02
Remove unused ignored_subst_vars argument from Deploy._subst_vars

rswarbrick:


machshev:
Yes this was part of https://github.com/lowRISC/dvsim/pull/35 and related work to refactor the config loading. It worked well and ran a lot faster than the current solution. However because of the attribute naming the default seed for example got eagerly substituted before the final seed value could be set.

So I parked the effort for the moment until we can start refactoring the HJSON in parallel, or migrate to a pydantic model of the HJSON and evaluate the HJSON wildcards more sensibly, i.e. by the parts of the system that know about them.

The learning I gained from this experiment was that a multi-pass wildcard evaluation was significantly faster than trying to evaluate everything all in one go in a single pass.

I'm currently working on the lint flow. Full rewrite with pydantic models, bypassing as much of the legacy code as I can. Almost done and I'll put it up for review.

rswarbrick:
Cool. In which case, let's drop this PR: I'm excited to see the nicer version! :-)

PR #158, opened 2026-04-02
Rewrite how OneShotCfg.build_modes gets constructed

rswarbrick:
This is a bit tricky because self.build_modes comes from parsing an hjson dictionary and is actually a list of dictionaries that describe build modes.

These get parsed and converted to BuildMode objects by Mode.create_modes.

I've written a rather silly loose check that makes sure the types look reasonable, which is enough to let us convince Pyright to allow the code to claim the correct eventual types.

This commit also fixes an incorrect type in the second argument to create_modes.


AlexJones0 (approved):
LGTM, modulo a few nits.

PR #160, opened 2026-04-02
Tidy up the last few Ruff problems in one_shot.py

rswarbrick:


src/dvsim/flow/one_shot.py
```diff
@@ -21,9 +22,25 @@
...
+    ignored_wildcards: ClassVar[list[str]] = [
+        *FlowCfg.ignored_wildcards,
+        "build_mode",
+        "index",
+        "test",
+    ]
 
-    def __init__(self, flow_cfg_file, hjson_data, args, mk_config) -> None:
+    def __init__(
+        self, flow_cfg_file: str, hjson_data: Mapping, args: object, mk_config: Callable
```
AlexJones0: Nit: more specific types here?

```suggestion
        self, flow_cfg_file: str, hjson_data: Mapping[str, Any], args: argparse.Namespace, mk_config: Callable[[str], FlowCfg]
```

PR #180, opened 2026-04-10
fix: stop stdout/stderr line buffering in local jobs when using `--interactive`

AlexJones0:
A follow-up fix to https://github.com/lowRISC/dvsim/pull/123.

That fix did solve the problem of log files not being captured for interactive jobs, but it introduced a new problem that the subprocess outputs were only displayed in a line-buffered fashion. This could cause problem for input prompts, e.g. where a subprocess asks "Continue [y/n] >" and expects an input, but via DVSim the user has not seen the output (due to no line break being seen).

Instead, expose a PTY interface and operate the tty in raw mode to remove all line buffering discipline and expose a terminal-like interface to the subprocess when running interactively. Wait for output via a selector in a loop, and use `os.read` on the file descriptors directly to handle the teeing of stdout/stderr to the log file. Keep stdin transparent to the subprocess itself however.

See the comments in the code for more details.

### DIFF
--- a/src/dvsim/runtime/local.py
+++ b/src/dvsim/runtime/local.py
@@ -6,10 +6,16 @@
 
 import asyncio
 import contextlib
+import os
+import pty
+import select
 import shlex
 import signal
 import subprocess
+import sys
+import termios
 import time
+import tty
 from collections.abc import Hashable, Iterable
 from dataclasses import dataclass
 from typing import TextIO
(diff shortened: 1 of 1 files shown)


machshev (approved):

AlexJones0:
Merging as the last force push was a comment-only change to fix a typo and CI is passing.

PR #182, opened 2026-04-10
Add Z01X sim tool runtime/simtime support

AlexJones0:
The format used for parsing the runtime and simulation time that is reported differs during the `run` phase for this simulation tool. Add the minimal relevant (somewhat fragile) logic to parse these values out so that they display correctly.

Also resolve some other miscellaneous cleanup surrounding the sim tools - see the commit messages for more information.


machshev (approved):
Thanks @AlexJones0

PR #183, opened 2026-04-16
ci: Use lowrisc/ci-actions for token acquisition

hcallahan-lowrisc:
Replace the local copy of lowrisc_ci_app_get_token with a reference to the canonical upstream action at lowrisc/ci-actions/get-token@v1. Also removes the preparatory checkout step that was only needed to make the local action available on disk.


AlexJones0 (approved):
LGTM, thanks @hcallahan-lowrisc.

PR #184, opened 2026-04-16
Augment the scheduler with resources to allow more fine-grained parallelism limitation

AlexJones0:
This PR contains the implementation of "resources" to the scheduler, which are essentially a mechanism for more fine-grained parallelism limits than what is currently offered by the standard `--max-parallel` flag. *Note*: This PR is quite large, let me know if it needs to be split up for review.

The `JobSpec` model is now changed so that each job can declare the resources that it uses, as a mapping of a resource name to some number. These resources are managed by a `ResourceManager` which is operated by the scheduler, which will strictly ensure that the running jobs do not try to allocate more resources than there are available. For now, resources are only defined for the various different sim tools, but this could be extended to the other flows in the future (and will probably be easier if the flows/deploys are better refactored).

Resources are determined via a `ResourceProvider` interface/protocol. Currently, this PR only implements static resources, where users can pass e.g. `--resource A=20 --resource B=50` flags on the command line to define static resource limits `{"A": 20, "B": 50}`. The main goal of this interface is that this can be extended in the future to support more dynamic resources if needed - for example license or compute availability that is actively polled via some command. While no dynamic resources are implemented in this PR, the integration of resources into the scheduler is designed such that no changes should be needed if/when they are introduced.

See the individual commit messages for more details. Also relevant: see my [local branch](https://github.com/AlexJones0/dvsim/commit/3a6266b0e48c1176b885b90b7320b398d66d9503) with some experimentation for dynamic resource availability.

### DIFF
--- a/src/dvsim/scheduler/resources.py
+++ b/src/dvsim/scheduler/resources.py
@@ -1,0 +1,203 @@
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+"""DVSim scheduler resource (parallelism limitation) management."""
+
+from collections import defaultdict
+from collections.abc import Iterable
+from enum import Enum
+from typing import Protocol
+
+from dvsim.job.data import JobSpec, ResourceMapping
+from dvsim.logging import log
+
+__all__ = (
+    "ResourceManager",
+    "ResourceProvider",
+    "StaticResourceProvider",
+)
+
+
+class ResourceProvider(Protocol):
+    """An abstraction of something that provides resource availability info to the scheduler."""
+
+    is_dynamic: bool
+    """Whether the resources returned by the provider can change over time."""
+
+    async def get_capacity(self) -> ResourceMapping:
+        """Get the capacity of resources available from this provider.
+
+        An integer capacity defines a strict limit for parallelism of that resource, whereas `None`
+        indicates no upper bound on parallelism.
+        """
+        ...
+
+
+class StaticResourceProvider:
+    """Provides information about static (unchanging) resource availability."""
+
+    is_dynamic = False
+
+    def __init__(self, limits: ResourceMapping) -> None:
+        """Construct a ResourceProvider with defined static limits."""
+        self._limits = limits
+
+    async def get_capacity(self) -> ResourceMapping:
+        """Get the static capacity of the resources available from this provider.
+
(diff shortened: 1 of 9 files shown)


src/dvsim/scheduler/resources.py
```diff
@@ -0,0 +1,202 @@
...
+            for resource, amount in request.items():
+                if amount is not None:
+                    self._usage[resource] += amount
+            return True
+
+    def release(self, request: ResourceMapping) -> None:
+        """Release a set of allocated resources."""
+        for resource, amount in request.items():
+            if amount is not None:
+                self._usage[resource] -= amount
```
machshev, line 110: Does this also need to be done within the lock?
It's still mutating the resource usage.

AlexJones0: No, releasing is monotonically decreasing and as such can never cause a problem; the lock isn't there to protect the `_usage` dictionary (like we would with a lock when threading), it's to stop `try_allocate` from seeing incorrect usages when awaiting `get_capacity` and thinking it can allocate when it cannot. That is, it's trying to stop problems from using two `try_allocate`s in a row. The lock is intentionally omitted from `release` for performance (both from not having the lock, but also meaning that `try_allocate` sees the released resources sooner and knows it can use them).

However, this actually made me consider this further, and I'm pretty sure the lock in `try_allocate` is also redundant and a bit useless. I originally added it whilst implementing to be on the safe side, but I don't think we can ever get in a bad state, because the there is no `await` between the check itself and the usage mutation like I thought might cause a problem. And thus no TOCTOU race in the async code. So I've just removed the lock instead.

Since this could potentially be a bit fragile to someone adding some more async logic, I've left a comment about this assumed invariant to hopefully avoid tripping anyone up!

AlexJones0:
Note: after some further discussion, I also dropped the [commit](https://github.com/lowRISC/dvsim/commit/9c78f099a4d1979633783f71542584eda65db6e7) that adds more granular per-tool license knowledge as part of the `SimTool` plugin. This might be nice eventually, but is a bit too complex at the moment and doesn't give us any useful advantages at the moment (all useful sub-licenses, e.g. those for formal, would all only be used for one job at once, because those use the GUI). When this PR is merged I'll create an issue to track this in case we might find it useful to add in the future.

Instead we just treat the tool *itself* as the resource. This has the nice advantage for now of meaning this generalizes and extends to support any other flows that also define a tool (e.g. linting? I think).

machshev (approved):
Thanks @AlexJones0!

PR #185, opened 2026-04-22
Action version bumps to resolve warnings about upcoming Node.js 20 EOL in CI

AlexJones0:
Bump all actions that were still using node20 to newer major versions that use node24. For each action I've checked the breaking changes and migration guidance in the release notes, and it appears that no breaking changes impact our use cases.

For any updated actions that support [immutable releases](https://docs.github.com/en/code-security/concepts/supply-chain-security/immutable-releases), I've pinned the specific immutable release. This would also mean pinning a specific SHA for the commonplace `checkout`, `upload-artifact` and `download-artifact` supported by GitHub, so for now these remain pinned only to major versions whilst no immutable releases exist. There are issues open on these actions upstream to support moving to immutable releases.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
       - uses: DeterminateSystems/flake-checker-action@main
 
       - name: Install uv
-        uses: astral-sh/setup-uv@v6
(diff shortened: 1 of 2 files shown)


hcallahan-lowrisc (approved):
LGTM! Thanks @AlexJones0

PR #186, opened 2026-04-29
feat(ci): enforce conventional commits

machshev:
We need to ensure that all commit messages are conventional commits as the auto release mechanism relies on this to work out what type of release to make. If it's a `fix:` then it will just do a patch release, `feat:` is a minor release, and breaking change is a major release.

Recently we had a commit go through that didn't have this format and it meant there was no release triggered.


martin-velay (approved):
That's a good idea to integrate this check into the CI for each commit. Thanks James

hcallahan-lowrisc:
As well as a CI check, we should add some docs to the contribution guidelines / a pre-populated PR message or some other sort of automated popup that reminds users they need to do this.

PR #187, opened 2026-04-30
Improve failure bucket presentation in block reports

AlexJones0:
Sort the buckets so the most common failure mode appears first, and also display some metadata listing the total number of failures in each bucket. This should make the generated block report more useful for finding important issues.

### DIFF
--- a/src/dvsim/templates/reports/block_report.html
+++ b/src/dvsim/templates/reports/block_report.html
@@ -11,7 +11,6 @@
 {% set timestamp = results.timestamp %}
 {% set stages = results.stages %}
 {% set coverage = results.coverage %}
-{% set failed_jobs = results.failed_jobs %}
 {% block content %}
 <div class="container-md">
     <div class="row">
@@ -204,7 +203,7 @@
     </div>
     {% endif %}
 
-    {% if failed_jobs.buckets %}
+    {% if failed_jobs %}
     <div class="row py-3">
         <div class="col">
             <h3>Error Messages</h3>
@@ -219,7 +218,7 @@
                     </tr>
                 </thead>
                 <tbody>
-                {% for msg, job_list in failed_jobs.buckets.items() %}
+                {% for msg, job_list in failed_jobs.items() %}
                     <tr class="table-primary">
                         <td colspan=5>{{ msg }}</td>
                     </tr>
--- a/src/dvsim/templates/reports/block_report.html
+++ b/src/dvsim/templates/reports/block_report.html
@@ -220,7 +220,8 @@
                 <tbody>
                 {% for msg, job_list in failed_jobs.items() %}
                     <tr class="table-primary">
-                        <td colspan=5>{{ msg }}</td>
+                        <td colspan=4>{{ msg }}</td>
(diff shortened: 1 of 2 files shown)


hcallahan-lowrisc (dismissed):

src/dvsim/templates/reports/block_report.html
```diff
@@ -219,7 +219,7 @@ <h3>Error Messages</h3>
                     </tr>
                 </thead>
                 <tbody>
-                {% for msg, job_list in failed_jobs.buckets.items() %}
+                {% for msg, job_list in failed_jobs.buckets.items() | sort(attribute='1|length', reverse=True) %}
```
rswarbrick: This is neat: I didn't know anything about Jinja before today!

How does the `attribute = '1|length'` bit work? I don't think I can find any examples like this when searching online.

AlexJones0: Thanks for pointing this out!

Jinja2 does support a [`dictsort`](https://jinja.palletsprojects.com/en/stable/templates/#jinja-filters.dictsort), but it doesn't seem to offer any way that I can tell to sort over the length of the values specifically.

Instead, when we call `.items()`, we are then iterating over the dictionary as a tuple of `(key, value)` pairs. Jinja2 supports the use of integer indexes (e.g. `1`) to sort over attributes of tuples. This works because for attribute accesses, [Jinja2 just follows the rule of regular `__getitem__` indexing](https://jinja.palletsprojects.com/en/stable/templates/#variables). So it relies on the fact that you can essentially do e.g. `next(iter(example_dict.items()))[1]` in Python to get the value. The `| length` mapping is then used to basically just call `len()` on the result. 

However, it turns out that actually this was not working! I had (incorrectly) assumed these can be chained together as in regular Jinja2 expressions, but it turns out the syntax in `attribute` here when sorting is not rich enough to express this sort of behaviour. And unfortunately it seems I was not rigorous enough in my testing, as the two cases I used to test already had the failures sorted, meaning I didn't catch it. I had assumed from the working test number annotations that things were working properly... apologies for that! I'm also not sure why Jinja2 isn't throwing an error in this case - it's obviously still interpreting it in some meaningful way.

I did come up with a solution to fix this in pure Jinja2, if you are interested:

```jinja2
{# We must sort within a namespace, as Jinja2's assignment scoping behaviour #}
{# prevents us from assigning within a loop. #}
{# See: https://jinja.palletsprojects.com/en/stable/templates/#assignments #}
{% set buckets = namespace(with_amounts=[]) %}
{% for k, v in failed_jobs.buckets.items() %}
    {% set buckets.with_amounts = buckets.with_amounts + [(v|length, k, v)] %}
{% endfor %}
{% for num_jobs, msg, job_list in buckets.with_amounts | sort(attribute=0, reverse=true) %}
```

However I think that this kind of Jinja2 logic is more complex than just pre-sorting the dictionary in Python before rendering the template, so I'm opting to switch to that solution instead.

hcallahan-lowrisc (approved):
Changed LGTM 👍 Thanks @AlexJones0

PR #188, opened 2026-04-30
Backend time/failure reporting fixes for timeouts of local jobs

AlexJones0:
This PR fixes two bugs in the runtime backends:
* One issue is inherited from the original launchers (whilst trying to maintain their behaviour), where any jobs that timeout do not report the DVSim-maintained runtime in the final results. We now always make sure to go through this path in the code, and make some refactoring improvements at the same time to centralize the timeout reporting logic to to reduce the potential for inconsistent failure messages in the future with more backends.
* The other is a small bug introduced in the `LocalRuntimeBackend`, where the error message that was being reported was not correct ("Job killed!" instead of "Job timed out after X minutes").

Example change:
```diff
 ### Test Results
 
 |  Stage  |  Name  | Tests                  |  Max Job Runtime  |  Simulated Time  |  Passing  |  Total  |  Pass Rate  |
 |:-------:|:------:|:-----------------------|:-----------------:|:----------------:|:---------:|:-------:|:-----------:|
-|         |        | ate_bootstrap_disjoint |      0.000s       |     0.000us      |     0     |    1    |   0.00 %    |
+|         |        | ate_bootstrap_disjoint |      6.007s       |     0.000us      |     0     |    1    |   0.00 %    |
 |         |        | **TOTAL**              |                   |                  |     0     |    1    |   0.00 %    |
 |         |        | **TOTAL**              |                   |                  |     0     |    1    |   0.00 %    |
 
 ## Coverage Results
 ### [Coverage Dashboard](/opentitan/scratch/master/chip_earlgrey_asic-sim-vcs/cov_report/dashboard.html)
 
 ## Failure Buckets
-* `Job killed!` has 1 failure:
+* `Job timed out after * minutes` has 1 failure:
     * Test ate_bootstrap_disjoint has 1 failure.
         * 0.ate_bootstrap_disjoint.58088541501576402788773390777979312745996416705647052408538425973305215300754\
           Log /opentitan/scratch/master/chip_earlgrey_asic-sim-vcs/0.ate_bootstrap_disjoint/latest/run.log
```

### DIFF
--- a/src/dvsim/runtime/backend.py
+++ b/src/dvsim/runtime/backend.py
@@ -218,11 +218,21 @@ class RuntimeBackend(ABC):
         self._make_job_output_directory(job)
 
     def _finish_job(
-        self, handle: JobHandle, exit_code: int, runtime: float | None
+        self, handle: JobHandle, exit_code: int | None, runtime: float | None
     ) -> tuple[JobStatus, JobStatusInfo | None]:
         """Determine the outcome of a job that ran to completion, and parse extra log info.
 
         Updates the handle with any extracted job runtime & simulation time info.
+
+        Args:
+            handle: The handle to the job that finished.
+            exit_code: The exit code if the job finished gracefully, or None if it was terminated.
+            runtime: The amount of time taken to complete the job, if known.
+
+        Returns:
+            A tuple containing the final job status and optionally an additional context/reason
+            object describing the job completion.
+
         """
         if handle.spec.dry_run:
             return JobStatus.PASSED, None
@@ -241,6 +251,11 @@ class RuntimeBackend(ABC):
             handle.simulated_time.set(*simulated_time.get())
 
         # Determine the final status from the logs and exit code.
+        if exit_code is None:
+            return JobStatus.KILLED, JobStatusInfo(
+                message=f"Job timed out after {handle.spec.timeout_mins} minutes"
+            )
+
         status, reason = log_results.get_status_from_logs()
         if status is not None:
             return status, reason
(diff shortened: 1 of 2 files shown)


src/dvsim/runtime/backend.py
```diff
@@ -218,11 +218,20 @@ def _prepare_launch(self, job: JobSpec) -> None:
...
 
         Updates the handle with any extracted job runtime & simulation time info.
+
+        Args:
+            handle: The handle to the job that finished.
+            exit_code: The exit code if the job finished gracefully, or None if it was terminated.
+            runtime: The amount of time taken to complete the job, if known.
+
+        Returns:
+            A tuple containing the final job status and an optional error reason (on failure).
```
hcallahan-lowrisc: Nit. Maybe `JobStatusInfo` needs a rename if it only captures context about errors/failures.

hcallahan-lowrisc (approved):
Makes sense to me. Thanks @AlexJones0

src/dvsim/runtime/backend.py
AlexJones0: Perhaps, but the intention behind the name was that it could be used to record reasons for other job statuses if we wanted to extend DVSim to be able to do that in the future. The type itself is not really "error" specific.

hcallahan-lowrisc: In that case the description should be more general I think. e.g.
```suggestion
            A tuple containing the final job status and optionally an additional context object about the job completion.
```

AlexJones0: That sounds like a good idea - I'll make that change now.

rswarbrick (approved):
Thanks for sorting this out.

The changes look right to me. For the "in reality..." comment on the second commit, should we add an issue tracking that this is worth doing?

AlexJones0:
> For the "in reality..." comment on the second commit, should we add an issue tracking that this is worth doing?

I think my wording here might have been a bit misleading - that comment is what the commit is doing! No more issue to be tracked :smiley:

PR #189, opened 2026-05-13
Refactor instrumentation interfaces to match the new scheduler

AlexJones0:
This PR is the first of a series of PRs to introduce instrumentation reporting to DVSim, which in this case also involves refactoring the existing instrumentation implementation to allow better modeling & static typing around this use case.

This PR performs some initial refactoring to the instrumentation logic to make improvements that are now possible due to the move to the new async scheduler interface. Notably:
* We remove the unnecessary `NoOpInstrumentation` - we can just not register callbacks instead.
* We replace the `CompositeInstrumentation` with an `instrumentationAggregator` interface - there is no need for arbitrary levels of nesting in instrumentation, and it will make introducing pydantic models later more difficult.
* We make use of the new `JobSpec.id` property introduced alongside the new scheduler rather than re-defining an unstable tuple on the job fields everywhere.

This should facilitate the movement to static/frozen Pydantic models for instrumentation data, which will mean we can define well-typed instrument report generation logic in later PRs. See the commit messages for more details.


machshev (approved):
Nice, Thanks @AlexJones0

PR #192, opened 2026-05-13
Convert instrumentation metric reporting to Pydantic models

AlexJones0:
This PR is the third of a series of PRs to introduce instrumentation reporting to DVSim, which in this case also involves refactoring the existing instrumentation implementation to allow better modeling & static typing around this use case.

This PR features the main overhaul to the instrumentation data model logic. Unfortunately this PR is on the slightly larger side by necessity, as it would otherwise require a lot of extraneous code & commits to make the migration cleanly. But the code should hopefully be relatively easy to review. The main changes involved are:
1. Replace all existing metric dataclasses with Pydantic models, and move them into a central `instrumentation/records.py` file. Stop re-exporting so many of these through the `instrumentation/__init__.py` package.
2. Add the ability to independently retrieve the scheduler & job metrics from each instrumentation.
3. Identify each instrumentation with a stable name attribute on the class itself.
4. Instead of collapsing the reported instrumentation metrics into some nested dictionary of arbitrary data, construct a Pydantic model for the combined instrumentation report. Note that this *does* change the generated output instrumentation logic slightly (but it should be an improvement in readability, in general).

For the 1st commit, I recommend reviewing `base.py`, then `records.py` alongside (`metadata.py`, `timing.py`, `resources.py`), and finally `__init__.py` for the easiest time.

This will allow us to introduce reporting logic later on which can consume the instrumentation data whilst being confident in the static typing provided. Plugins can then still be supported via Pydantic's `extra=allow` config on the combined scheduler and job instrumentation data models.

### DIFF
--- a/src/dvsim/instrumentation/__init__.py
+++ b/src/dvsim/instrumentation/__init__.py
@@ -6,13 +6,19 @@
 
 from dvsim.instrumentation.base import InstrumentationAggregator, SchedulerInstrumentation
 from dvsim.instrumentation.factory import InstrumentationFactory
-from dvsim.instrumentation.records import InstrumentationMetrics, JobMetrics, SchedulerMetrics
+from dvsim.instrumentation.records import (
+    InstrumentationMetrics,
+    InstrumentationResults,
+    JobMetrics,
+    SchedulerMetrics,
+)
 from dvsim.instrumentation.runtime import flush, get, set_instrumentation, set_report_path
 
 __all__ = (
     "InstrumentationAggregator",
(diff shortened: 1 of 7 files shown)


machshev (approved):
Thanks @AlexJones0! This looks like a nice tidy up...

AlexJones0:
The latest force push is a rebase on `master` now that #189 has been merged, with no other changes.

AlexJones0:
And the latest force-push just fixed a bad copy-paste in a docstring :sweat_smile:

PR #193, opened 2026-05-13
Instrumentation refactoring cleanups & UX improvements

AlexJones0:
This PR is the fourth of a series of PRs to introduce instrumentation reporting to DVSim, which in this case also involves refactoring the existing instrumentation implementation to allow better modeling & static typing around this use case.

This PR performs some final small cleanups after the instrumentation refactoring work in the preceding PRs. Alongside removing some duplication of the instrumentation names, it primarily focuses on improving the user experience:
1. Make metadata an explicit instrumentation type, to replace the unintuitive behaviour where it is automatically added alongside other instrumentation types.
2. Rename the `resources` instrumentation to `compute`, to make it clear that we are measuring "compute resources" and not the new "resources" concept that was introduced to the scheduler in #184 for more granular parallelism limits.
3. Add a special `all` instrumentation type which automatically includes all the available instrumentations. This lets you type `--instrument all` instead of `--instrument meta timing compute`, which can get overly verbose.

### DIFF
--- a/src/dvsim/cli/run.py
+++ b/src/dvsim/cli/run.py
@@ -833,7 +833,7 @@ def parse_args(argv: list[str] | None = None):
         dest="instrumentation",
         nargs="+",
         default=[],
-        choices=InstrumentationFactory.options(),
+        choices=["all", *InstrumentationFactory.options()],
         help="Enable scheduler instrumentation (can specify multiple types).",
     )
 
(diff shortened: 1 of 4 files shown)


machshev (approved):
Nice, thanks @AlexJones0!

AlexJones0:
The latest force push is a rebase on `master` now that #189 and #192 have been merged, with no other changes.

PR #194, opened 2026-05-14
Add block information to the instrumentation job metadata

AlexJones0:
This PR is the fifth of a series of PRs to introduce instrumentation reporting to DVSim.

A very simple addition - also capture the block & block variant alongside the existing metadata during instrumentation. Although this could technically be inferred from the full name of the job, having these fields explicitly available through a clear interface is much cleaner.

This will allow us to make visualizations / calculate metrics based on run results partitioned by block or block variant.

### DIFF
--- a/src/dvsim/instrumentation/metadata.py
+++ b/src/dvsim/instrumentation/metadata.py
@@ -41,6 +41,8 @@ class MetadataInstrumentation(SchedulerInstrumentation):
                 job_type=spec.job_type,
                 target=spec.target,
                 tool=spec.tool.name,
+                block=spec.block.name,
+                block_variant=spec.block.variant,
                 backend=spec.backend,
                 dependencies=list(spec.dependencies),
                 status=status_str,
--- a/src/dvsim/instrumentation/records.py
+++ b/src/dvsim/instrumentation/records.py
@@ -52,6 +52,8 @@ class JobInstrumentationMetadata(JobMetrics):
     job_type: str
     target: str
     tool: str
+    block: str
+    block_variant: str | None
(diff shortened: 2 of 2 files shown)


src/dvsim/instrumentation/metadata.py
```diff
@@ -41,6 +41,8 @@ def get_job_data(self) -> Mapping[str, JobInstrumentationMetadata]:
                 job_type=spec.job_type,
                 target=spec.target,
                 tool=spec.tool.name,
+                block=spec.block.name,
```
machshev, line 44: Maybe rather than flattening just use composition and take the whole block meta data `spec.block`... that way if we update the block definition in some other way then this is constant without having to sync everything.

It does mean more information though and increased serialisation size perhaps?

machshev (approved):
Looks good, though I might prefer adding the whole block object (which is supposed to be block identifying meta data). At the moment it's maybe a bit heavy. But longer term I think we could deduplicate and move derived names to calculated properties. Which minimises serialised data size.

src/dvsim/instrumentation/metadata.py
AlexJones0: Yeah I agree, I'm conscious that any fields added here directly impact the instrumentation size by a fair amount, so I'm trying to avoid bloating the output too much.

I think long-term we should split up the block info (`IPMeta`) a bit more to separate block name info from commit/revision info, but I suspect that will be a slightly larger task due to the number of involved areas, and so I'd prefer if that happened after this instrumentation reporting work was merged. WDYT?

PR #196, opened 2026-05-14
Add HTML instrumentation report types, registry & template generation

AlexJones0:
This PR is the sixth of a series of PRs to introduce instrumentation reporting to DVSim.

This PR adds an initial basis for the instrumentation HTML report generation. It features:
* The HTML template for the instrumentation report, including the introduction of relevant dependencies on plotly (will be used by future PRs which add visualizations to the report).
* A protocol / abstraction for how instrumentation report visualizations should be implemented, and a registry to register these visualizations (this can let end users hook in with custom instrumentation reporting logic for their custom instrumentation implementations).
* The bridge/glue logic to generate the HTML instrumentation report after DVSim runs with instrumentation enabled.

A few important things to note:
* Note that the visualization rendering path of the HTML is not actually used yet, as no visualizations have been implemented. These are intended to be added in future PRs, to ease the review burden.
* The minfied plotly.js dependency comes built into the `plotly` python package (so we don't need to worry about vendoring & versions becoming out of sync). We only generate the static `plotly.min.js` file if there are any visualizations available to actually use plotly.
* For now, whilst the report is still WIP across this set of PRs, we do not link to the generated `metrics.html` anywhere from the main sim flow `index.html` page. This will be added later.

See the commit messages for more info & context about each change.

---

Currently, the generated `metrics.html` report will look something like:

Screenshot from 2026-05-14 12-36-53

With the `JSON` button taking you to the existing `metrics.json` file.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,6 +23,7 @@ dependencies = [
     "gitpython>=3.1.45",
     "hjson>=3.1.0",
     "jinja2>=3.1.6",
+    "plotly>=6.7.0",
     "psutil>=7.2.2",
     "pydantic>=2.9.2",
     "pyyaml>=6.0.2",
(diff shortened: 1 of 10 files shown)


machshev (approved):
Nice, thanks @AlexJones0

PR #197, opened 2026-05-14
Introduce instrumentation report utilities

AlexJones0:
**Note: this PR is marked as a draft as it currently depends on #194 and #196. The first commit of this PR is from #194, and the 2nd-7th commits are from that #196. These can be safely ignored - only the 8th-11th commits are relevant to this PR. When #194 and #196 are merged, this PR will be updated and marked as ready to review, but is otherwise ready for review.**

This PR is the seventh of a series of PRs to introduce instrumentation reporting to DVSim.

This PR adds some utility functions & methods that will be used throughout the various instrumentation visualizations/graphs. These are not used yet, in an effort to split the changes up into multiple PRs that should be easier to review. Changes include:
* Time string formatting not already available through libraries like `datetime`.
* Add some helper methods to the instrumentation report to get concrete timing information where only partial info may exist (as far as the existing typed models go).
* Helper function for automatically rendering a plotly Figure as a PNG instead of HTML when a threshold of entities in the figure is exceeded. This supports scaling in a generic way so that we don't get 250MB+ reports for large runs :)
* Helper functions for making job metadata and colour mappings to use for plotly figures, to ensure consistent presentation across various visualisations.


AlexJones0:
The licensecheck issue causing a CI failure is a minor issue that can be easily worked around (the `licensecheck` tool doesn't handle licenses with exceptions well). But it has exposed to me some issues with the underlying approach, so I need to spend a bit of time redoing this.

Specifically:
* For two of the visualizations that will be added (Gantt chart & parallelism chart), these grow very large for large runs (e.g. 50k+ jobs).
* This both bloats report file size (e.g. from 7 MiB to 250+ MiB for ~50K jobs) and drastically hurts performance when trying to browse the graph.
* My original solution: since you mostly care about the shape at this scale and interactivity isn't important (with the normal rendering profile), just export it as a PNG. This was working well for me locally.

However it turns out that since Plotly (and many other big Python visualization frameworks) are based in JS, they need a browser to render before being converted to an image. Even worse, Plotly relies on the `kaleido`/`choreographer` dependencies to do this, which have a strict dependency on Chrome being available on the system.

Rather annoyingly, although [Kaleido's README.md](https://github.com/plotly/Kaleido#installation) mentions the requirement for Chromium, the [Plotly README.md](https://github.com/plotly/plotly.py#kaleido) says:

> The [kaleido](https://github.com/plotly/Kaleido) package has no dependencies and can be installed using pip
    
(which feels a little misleading).

I'd really rather not make DVSim depend on a specific browser being installed if possible. One solution would be to just not render these graphs at scale, but another solution I'm exploring which looks quite feasible is just to render these graphs in `matplotlib` over a certain scale instead. Matplotlib can render directly to an image without going through a browser, and it doesn't seem *too* hard to convert between the two frameworks, especially since both problematic graphs actually just use the same underlying implementation, so it's really only translating one graph. The main point of difficulty will be abstracting in a nice way to not duplicate a lot of stuff between the two implementations.

Since I'm exploring that avenue and I don't intend to add the `kaleido` dependency nor use my original `render_large_image` helper, I'm closing this PR for now.

PR #198, opened 2026-05-19
fix: Raise a helpful error on a bogus testplan

rswarbrick:
I found this because I had added a problem to a testplan and running dvsim on the resulting block didn't work. But it failed silently!

Now, you get a backtrace:
```
  ...
    File "/home/rjs/work/dvsim/src/dvsim/sim/flow.py", line 324, in _create_objects
      self.testplan = Testplan(
                      ^^^^^^^^^
    File "/home/rjs/work/dvsim/src/dvsim/testplan.py", line 350, in __init__
      self._parse_testplan(filename, tags, repo_top)
    File "/home/rjs/work/dvsim/src/dvsim/testplan.py", line 463, in _parse_testplan
      self.covergroups = self._create_testplan_elements("covergroup", covergroups, set())
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/home/rjs/work/dvsim/src/dvsim/testplan.py", line 284, in _create_testplan_elements
      raise ValueError(err_msg)
  ValueError: Duplicate items with name handshake_complete_cg.

(which also tells me what I did wrong in my OpenTitan patch!)
```

### DIFF
--- a/src/dvsim/testplan.py
+++ b/src/dvsim/testplan.py
@@ -255,24 +255,31 @@ class Testplan:
         sys.exit(1)
 
     @staticmethod
-    def _create_testplan_elements(kind: str, raw_dicts_list: list, tags: set):
-        """Creates testplan elements from the list of raw dicts.
+    def _create_testplan_elements(kind: str, raw_dicts_list: list, tags: set) -> list[Element]:
+        """Create testplan elements from the list of raw dicts.
 
         kind is either 'testpoint' or 'covergroup'.
         raw_dicts_list is a list of dictionaries extracted from the HJson file.
         """
         items = []
         item_names = set()
+
+        try:
+            elem_cls = Testplan.element_cls[kind]
+        except KeyError:
+            err_msg = f"No known class for kind={kind}"
+            raise RuntimeError(err_msg) from None
+
         for dict_entry in raw_dicts_list:
             try:
-                item = Testplan.element_cls[kind](dict_entry)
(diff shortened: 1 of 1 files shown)


src/dvsim/testplan.py
```diff
@@ -255,24 +255,31 @@ def _parse_hjson(filename):
...
+        """Create testplan elements from the list of raw dicts.
 
         kind is either 'testpoint' or 'covergroup'.
         raw_dicts_list is a list of dictionaries extracted from the HJson file.
         """
         items = []
         item_names = set()
+
+        try:
+            elt_cls = Testplan.element_cls[kind]
```
AlexJones0: I've not seen `elt` as a shorthand for `element` before. Maybe `elem` or `elmt` would be more appropriate?

AlexJones0 (approved):
Thanks, this looks like a big improvement to me!

AFAIK the testplan code has barely been touched during refactoring, but it remains quite integrated into the various flows - there are probably many things that can be improved upon in this file.

machshev (approved):

src/dvsim/testplan.py
rswarbrick: I think this was a shorthand of maths professors when I was an undergraduate :-) I didn't realise it was so non-standard!

I'll switch to `elem` now.

PR #200, opened 2026-05-20
Introduce instrumentation report utilities

AlexJones0:
This PR is the eighth of a series of PRs to introduce instrumentation reporting to DVSim.

This PR adds some helper utilities that will be used throughout the various instrumentation visualizations. These are not used yet in this PR - in an effort to split the changes up into multiple PRs and make them easier to review. There are a variety of helpers across the commits - some for time string formatting, some for retrieving concrete timing information from the report data, and some for rendering Plotly figures and deriving consistent colour mappings across visualizations.

(note on the time string formatting specifically: `datetime` does let you get some format like "Xh Ym Zs", but it doesn't let you omit leading zero values which is nice for presentation in the reports. We could hack on top and strip leading string prefixes but it's simple enough to just re-implement that I decided to do that here. We also could pull in some third-party deps, but that seems unnecessary for just this functionality).

### DIFF
--- a/src/dvsim/instrumentation/report/base.py
+++ b/src/dvsim/instrumentation/report/base.py
@@ -4,24 +4,44 @@
 
 """DVSim scheduler instrumentation reporting & visualizations."""
 
-from collections.abc import Sequence
+from collections.abc import Iterable, Mapping, Sequence
 from enum import Enum
 from pathlib import Path
-from typing import Protocol
+from typing import Any, Protocol, TypeVar
 
+import plotly.graph_objects as go
 import plotly.offline
 from typing_extensions import Self
 
 from dvsim.instrumentation import InstrumentationResults
+from dvsim.instrumentation.records import JobInstrumentationMetadata
 from dvsim.logging import log
 from dvsim.report.artifacts import ReportArtifacts, render_static_content
 from dvsim.templates.render import render_template
 
 __all__ = (
+    "DEFAULT_VISUALIZATION_HEIGHT_PX",
+    "PLOTLY_TIMING_AXIS_CONFIG",
     "InstrumentationVisualizer",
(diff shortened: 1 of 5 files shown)


src/dvsim/utils/time.py
```diff
@@ -12,3 +12,44 @@
 
 # Timestamp format for Hours:Minutes:Seconds display as used by the scheduler
 TS_HMS_FORMAT = "%H:%M:%S"
+
+
+def format_time_as_hms(seconds: float, *, decimals: int = 2, omit_zero: bool = False) -> str:
```
machshev, line 17: Looks good but I'd probably add some parameterize unit tests for this to catch corner cases or in case someone updates the implementation and breaks it?

src/dvsim/instrumentation/report/base.py
```diff
@@ -130,3 +150,90 @@ def render_html_report(
             (outdir / plotly_js_path).write_text(artifacts[plotly_js_path])
 
     return artifacts
+
+
+def render_plotly_figure(fig: go.Figure, *args: list[Any], **kwargs: dict[str, Any]) -> str:
```
machshev, line 155: `go.Figure` ;)

machshev (approved):
Thanks @AlexJones0

src/dvsim/instrumentation/report/base.py
AlexJones0: Several times I've found this amusing whilst working on this code :)

src/dvsim/utils/time.py
AlexJones0: Great point, I didn't think of this at all. I'll add it now.

AlexJones0: Added.

PR #201, opened 2026-05-20
Add job timeline Gantt chart to the HTML instrumentation report

AlexJones0:
*Note: most of the diff is a lockfile change, the actual size of this PR is just under +400 lines or so).*

This PR is the ninth of a series of PRs to introduce instrumentation reporting to DVSim.

This PR introduces the first of the instrumentation report visualizations - a Gantt chart showing a timeline of all the jobs scheduled by a DVSim run. This graph gives some overall context for a run and can help highlight some potential long-running jobs and tail domination at a glance.

Much more detail is in the commit messages - I highly recommend giving those a read for motivations behind some of the decisions.

---

* A Plotly graph generated for an OpenTitan "nightly" run:
image

* A matplotlib graph generated for an OpenTitan "weekly" run:
image

* A Plotly graph for a local run of some OpenTitan uart smoketests:
image

* An example of the tooltip on hovering over a job (where metadata is available):
image


src/dvsim/instrumentation/report/timelines.py
```diff
@@ -182,3 +218,144 @@ def render(self, results: InstrumentationResults) -> str | None:
...
+        fig.savefig(buf, format="png", dpi=fig_dpi)
+        png_bytes = buf.getvalue()
+        plt.close(fig)
+
+        # For now, to keep this abstraction simple, encode the image as a base64 png.
+        # TODO: it might be nice to render this as a separate report artifact in the future, but
+        # the `InstrumentationVisualizer` interface will need some work.
+        b64 = base64.b64encode(png_bytes).decode("ascii")
+        return (
+            f'<img src="data:image/png;base64,{b64}" '
```
machshev, line 315: This could result in quite big images, separate image file and a relative link would be better. But this works for now.

machshev (approved):
Thanks @AlexJones0

src/dvsim/instrumentation/report/timelines.py
AlexJones0: Agreed - in practice I've tried running it with both real and synthetic data and found the following:
* At < 50k jobs, the generated report size (with just this graph embedded) will vary based on compression etc. but tends to sit at around < 0.4 MB.
* At 250k jobs, the generated report size is around 0.45 MB.

So perhaps a little larger than it could be if we re-architect to save directly as a PNG and load with a relative link, but small enough that I'm comfortable making this workaround.

I think it might be best to open an issue related to addressing this, rather than spending more time on it now?

machshev: Yes makes perfect sense...

AlexJones0: I've created an issue to track this here: https://github.com/lowRISC/dvsim/issues/203

PR #204, opened 2026-05-20
Add graph of tool usage over time to the HTML instrumentation report

AlexJones0:
This PR is the eleventh of a series of PRs to introduce instrumentation reporting to DVSim.

This PR adds a 3rd instrumentation report visualization, which is a time series graph showing how tools are used over time. This graph is only created if metadata is available in the instrumentation report. At each point in time, the graph should visualize how many jobs were running that use a given tool. The motivation is to expose:
* The scope of usage of each tool.
* The parallelism of using each tool.
* Potential limitations in overall parallelism and the need to apply more granular parallelism limits through e.g. the resource mechanism.

---

* A graph generated for an OpenTitan "nightly" run:
image

* A graph generated for an OpenTitan "weekly" run:
image


machshev (approved):
Nice, thanks @AlexJones0!

PR #205, opened 2026-05-20
Add tool usage use and block breakdown charts to the HTML instrumentation report

AlexJones0:
This PR is the twelfth of a series of PRs to introduce instrumentation reporting to DVSim.

This PR adds two more visualizations to the instrumentation report, both of which are "breakdown" visualizations. These partition the jobs that ran into groups based on available metadata, sum up the duration of jobs in each of these groups, and present this info in a structured manner via a pie chart and bar graph. Breakdowns are specifically added for tool usage and for block (variant) scopes, to allow identification of the (a) scope of usage of each tool and (b) relative contributions of each block to the overall run duration.

*Note*: for the breakdown chart, you can click on different traces in the legend to remove or add certain groups to the bar chart. Ideally I would like this to affect both the pie chart and bar chart simultaneously to keep the view consistent, and indeed Plotly has a "legendgroup" feature that can be used to do this - but it does not work if one of the subplot figures is a Pie chart, as Pie charts in Plotly are always modeled as having a single "trace" (unlike e.g. bar charts).

---

* A per-tool breakdown graph generated for an OpenTitan "weekly" run:
image

* A per-block breakdown graph generated for an OpenTitan "weekly" run:
dvsim_eg_weekly_blocks

### DIFF
--- a/src/dvsim/instrumentation/report/breakdown.py
+++ b/src/dvsim/instrumentation/report/breakdown.py
@@ -1,0 +1,227 @@
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+"""DVSim scheduler instrumentation breakdown (pie & bar chart) visualizations."""
+
+from collections import defaultdict
+from collections.abc import Callable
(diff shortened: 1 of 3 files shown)


machshev (approved):
Thanks @AlexJones0

PR #206, opened 2026-05-20
Add chart ranking the longest jobs per status to the HTML instrumentation report

AlexJones0:
This PR is the thirteenth of a series of PRs to introduce instrumentation reporting to DVSim.

This PR introduces a new visualization type which ranks the longest jobs, either globally or broken-down by status (`Passed`, `Failed`, and `Killed`). The motivation is to let you easily be able to determine the largest contributors to your overall run duration, and to determine jobs that might be failing or timing out (and taking a significant time whilst doing so). Buttons are provided for filtering between different views (combined or per-status) to allow more controlled browsing of the data.

---

* The combined "All categories" view generated for an OpenTitan "nightly" run:
image

* The filtered "Passing" view generated for an OpenTitan "weekly" run:
image


machshev (approved):
Thanks @AlexJones0!

PR #207, opened 2026-05-20
Add charts ranking the longest jobs & tests per tool to the HTML instrumentation report

AlexJones0:
This PR is the fourteenth of a series of PRs to introduce instrumentation reporting to DVSim.

This PR extends the visualization introduced in the previous PR to rank the longest jobs so that it can now rank the longest "tests" as well - where a test is a group of jobs that are running the same thing, just with different seeds. It then adds two new visualizations to the HTML instrumentation report: one ranking the top jobs partitioned by tool, and another ranking the top tests partitioned by tool.

Three key areas of note / questions:
1. For the "test" view, jobs are presented in a stacked fashion using Plotly's "overlay" mode so that we can still break the test run down into its individual jobs. This is a really cool feature... but it doesn't quite scale well to the 10k+ jobs we might want to use it for. To save performance here we only individually render the top 5 jobs per test by default, and combine the others into a single bar with aggregate info. The rendering profiles still provide a mechanism for users to retrieve more/full detail if desired, to try and keep the the default small & performant.<br></br>
2. There is some unfortunately hacky code in `LongestBarChart.get_job_test` to determine the test for a given job, so that we can group similar jobs together. It uses the presence of a seed to determine if something is a test (otherwise many "default" `build`, `cov_merge` and `cov_report` jobs are combined together), and if so uniquely identifies it by the test name and the block variant combo. In my opinion, this concept needs to be better integrated into the `JobSpec` model itself - we already have some confusion around `name` vs. `qual_name` vs. `full_name` but in this case, you can see that the granular information needed to determine test groupings is not sufficiently available (as far as I can tell), so I have to resort to this.<br></br>
3. Related to the above, in terms of naming, I *think* that "tests" is an appropriate way to refer to a "group of jobs that are basically running the same thing, with different seeds", but I'm not sure how much that concept is actually enforced in DVSim. If DVSim was made sufficiently generic in the future, this name might no longer hold? Perhaps that's a discussion for refactoring around the `FlowCfg`, `Testplan` & `Deploy` abstractions?

---

* The "xcelium" filtered view of the top jobs by tool, generated for an OpenTitan "weekly" run:
image

* The combined "All categories" view of the top tests by tool, generated for an OpenTitan "weekly" run:
image

* An example of the tooltip text on hover for a combined job bar in the "top tests by tool" view, generated for an OpenTitan "weekly" run:
image

### DIFF
--- a/src/dvsim/instrumentation/report/longest.py
+++ b/src/dvsim/instrumentation/report/longest.py
@@ -2,10 +2,11 @@
 # Licensed under the Apache License, Version 2.0, see LICENSE for details.
 # SPDX-License-Identifier: Apache-2.0
 
-"""DVSim scheduler instrumentation bar chart visualizations of the longest jobs."""
+"""DVSim scheduler instrumentation bar chart visualizations of the longest jobs & tests."""
 
 from collections import defaultdict
 from collections.abc import Callable
+from dataclasses import dataclass
 from typing import Any
 
 import plotly.colors as pc
@@ -15,6 +16,7 @@ from typing_extensions import Self
 from dvsim.instrumentation import InstrumentationResults
 from dvsim.instrumentation.records import (
     ConcreteJobTimingMetrics,
+    JobInstrumentationMetadata,
     JobInstrumentationResults,
 )
 from dvsim.instrumentation.report.base import (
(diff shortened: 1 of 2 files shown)


machshev:
> For the "test" view, jobs are presented in a stacked fashion using Plotly's "overlay" mode so that we can still break the test run down into its individual jobs. This is a really cool feature... but it doesn't quite scale well to the 10k+ jobs we might want to use it for.

If you want to show variation within a job then maybe a candlestick chart showing the distributions of durations per job as a separate chart? This would summarise the information, variation within a test?

machshev (approved):
Thanks @AlexJones0

AlexJones0:
> If you want to show variation within a job then maybe a candlestick chart showing the distributions of durations per job as a separate chart? This would summarise the information, variation within a test?

This is quite a nice idea! So I guess we still have the same kind of view where we rank tests by their duration, but we'd instead be showing the distributions of job durations for each test? I guess the one downside compared to the current approach is that you lose the ability to hover and inspect the info (e.g. seed number) for individual job runs, but I'd say presentation-wise what you're suggesting might be easy to visually parse.

PR #208, opened 2026-05-20
Add charts ranking the longest jobs & tests per block (variant) to the HTML instrumentation report

AlexJones0:
This PR is the fifteenth of a series of PRs to introduce instrumentation reporting to DVSim.

This PR adds the final two remaining visualizations for the HTML instrumentation report implementation: one ranking the top jobs partitioned by block (variant), and another ranking the top tests partitioned by block (variant). Compared to the previous PRs, this should be a lot nicer to review - we're just extending the functionality introduced in the last two PRs via parameterization.

---

* The "otbn" filtered view of the top jobs by block, generated for an OpenTitan "nightly" run:
image

* The combined "All categories" view of the top tests by tool, generated for an OpenTitan "weekly" run:
image

### DIFF
--- a/src/dvsim/instrumentation/report/longest.py
+++ b/src/dvsim/instrumentation/report/longest.py
@@ -597,3 +597,93 @@ class LongestTestsByToolChart(LongestByToolChart):
             )
             return cls(max_bars=None, max_jobs_per_bar=None)
         return cls()
+
+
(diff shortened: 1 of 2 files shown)


machshev (approved):
Thanks @AlexJones0!

PR #209, opened 2026-05-20
Make colour handling more consistent in the HTML instrumentation report

AlexJones0:
This PR is the sixteenth of a series of PRs to introduce instrumentation reporting to DVSim.

This PR does a little bit of clean up on the instrumentation reports to make the handling of colours in the plots more consistent. First, we turn the colour mapping functionality implemented for the breakdown visualizations into a general utility that is used by all the report types, as this is useful for the edge cases where we have a lot more categories/traces than Plotly's default colours (of which there are 10). Since e.g. the charts breaking down jobs/tests by block will probably need more than that, we inevitably see repeating colours - using a larger default colour set can avoid this. The second commit is a minor fix to said "longest N jobs/tests" chart(s) to make the colour mapping consistent with the other generated plots.

### DIFF
--- a/src/dvsim/instrumentation/report/base.py
+++ b/src/dvsim/instrumentation/report/base.py
@@ -9,6 +9,7 @@ from enum import Enum
 from pathlib import Path
 from typing import Any, Protocol, TypeVar
 
+import plotly.colors as pc
 import plotly.graph_objects as go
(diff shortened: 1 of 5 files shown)


machshev (approved):
Nice, thanks @AlexJones0!

PR #210, opened 2026-05-20
Add a link to the instrumentation report from the sim flow HTML summary report

AlexJones0:
This PR is the seventeenth of a series of PRs to introduce instrumentation reporting to DVSim.

This is intended to be merged after the visualizations have been added via the other instrumentation report PRs, but it doesn't strictly depend on them. If DVSim is run with instrumentation enabled and a report was generated, then this will place a link to that instrumentation report ("metrics.html") in the generated simulation flow report summary. The idea is to make sure that we have one central index for simulation flows from which all the generated report data can be reached.

---

Now, the generated simulation `index.html` summary report will look something like:

image

With the `Instrumentation Report` button taking you to the `metrics.html` report.

### DIFF
--- a/src/dvsim/sim/report.py
+++ b/src/dvsim/sim/report.py
@@ -125,13 +125,20 @@ class HtmlReportRenderer:
             if outdir is not None:
                 (outdir / block_file).write_text(artifacts[block_file])
 
+        # If instrumentation report data is available, we should also link to that generated report
+        instrumentation_report = None
+        if outdir is not None:
+            instrumentation_path = outdir / "metrics.html"
+            if instrumentation_path.exists() and instrumentation_path.is_file():
+                instrumentation_report = "metrics.html"
+
         # Regardless of whether we have a top or there is only one block, we always generate a
         # summary page for now.
         top_log_suffix = "" if summary.top is None else f" for {summary.top.name}"
         log.debug("Generating HTML summary report%s", top_log_suffix)
         artifacts["index.html"] = render_template(
             path="reports/summary_report.html",
-            data={"summary": summary},
+            data={"summary": summary, "instrumentation_report": instrumentation_report},
         )
         if outdir is not None:
             (outdir / "index.html").write_text(artifacts["index.html"])
--- a/src/dvsim/templates/reports/summary_report.html
+++ b/src/dvsim/templates/reports/summary_report.html
@@ -37,9 +37,9 @@
         </div>
     </div>
 
-    <div class="row">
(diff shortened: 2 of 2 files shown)


src/dvsim/sim/report.py
```diff
@@ -125,13 +126,16 @@ def render(
             if outdir is not None:
                 (outdir / block_file).write_text(artifacts[block_file])
 
+        # If instrumentation report data is available, we should also link to that generated report
+        instrumentation_report = None if instrumentation.get_report() is None else "metrics.html"
```
machshev: I'd probably just use `pathlib` to check to see if the file exists. Which means it works if the report is being regenerated or not?

src/dvsim/templates/reports/summary_report.html
```diff
@@ -64,6 +64,14 @@ <h2>{{ title }}</h2>
             </span>
             {% endif %}
         </div>
+        {% if instrumentation_report %}
+        <div class="col-4 col-md-3">
+            <a class="badge text-bg-secondary link-underline link-underline-opacity-0"
+                style="font-size: 15px;" href="{{ instrumentation_report }}">
```
machshev: Preferably use bootstrap classes here instead of custom CSS. There are font size classes (e.g. `fs-6`) https://getbootstrap.com/docs/5.3/utilities/text/#font-size and https://getbootstrap.com/docs/5.3/content/typography/#inline-text-elements shows there is a `small`.

If nothing fits then custom CSS is okay, it's just it's neater if you can describe what you want as a class. Which means that if the template gets updated then things scale functionally.

machshev (approved):
Thanks @AlexJones0!

Couple of nits but not major.

src/dvsim/templates/reports/summary_report.html
AlexJones0: I haven't done too much web stuff before so I wasn't aware that this was available. I'll look into changing this. Thanks for catching it!

AlexJones0: Changed this now - `fs-6` was exactly what I needed.

src/dvsim/sim/report.py
AlexJones0: Sure - I've had to derive the path from the given `outdir` - which means that if your intention isn't to immediately write the output but instead to generate some report artifacts for manipulating in Python this flow still isn't perfect - but it's better than what I had before.

machshev (approved):
Nice, thanks @AlexJones0!

PR #211, opened 2026-05-20
Add instrumentation report (re-)generation to the `dvsim-admin` CLI

AlexJones0:
This PR is the eighteenth of a series of PRs to introduce instrumentation reporting to DVSim.

This PR integrates the HTML instrumentation report generation with the `dvsim-admin` CLI to allow reports to be re-generated from previously saved `metrics.json` files after DVSim runs have already completed.

Importantly, it also provides an interface for specifying a custom rendering profile. This means that we can render with the "Normal" profile by default for regular DVSim runs, minimizing the size and increasing the performance of the generated instrumentation reports - whilst at the same time retaining the ability to browse fully-detailed reports via the `dvsim-admin` CLI.

For example, you might run:
```
dvsim <your arguments here> --instrument all
```
And then later run e.g.
```
dvsim-admin report instrumentation --profile=full scratch/<branch>/reports/metrics.json .
```
to generate a new `metrics.json` with fully-detailed graphs (without any regard for report size or performance).

### DIFF
--- a/src/dvsim/cli/admin.py
+++ b/src/dvsim/cli/admin.py
@@ -62,7 +62,7 @@ def dashboard_gen(json_path: Path, output_dir: Path, base_url: str | None) -> No
 
 @cli.group()
 def report() -> None:
-    """Reporting helper commands."""
+    """Reporting helper commands."""  # noqa: D401
 
 
 @report.command("gen")
--- a/src/dvsim/cli/admin.py
+++ b/src/dvsim/cli/admin.py
@@ -10,6 +10,8 @@ from pathlib import Path
 
 import click
 
+from dvsim.instrumentation.report.profile import RenderProfile
+
 
 @click.group()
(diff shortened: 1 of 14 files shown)


src/dvsim/cli/admin.py
```diff
@@ -10,6 +10,8 @@
 
 import click
 
+from dvsim.instrumentation.report import RenderProfile
```
machshev: Might be worth just checking how much this effects the click shell autocompletions?

https://click.palletsprojects.com/en/stable/shell-completion/#enabling-completion

In the past I've found that depending on what is pulled in by imports like this it can increase auto completion times by seconds... making them unusable. This is the admin cli so maybe it's not that much of an issue?

machshev (approved):
Thanks @AlexJones0!

machshev (approved):
Thanks @AlexJones0!

PR #214, opened 2026-05-21
doc: Add documentation to the Result class in testplan.py

rswarbrick:
In particular, this gives types to all the arguments and gives units for the numbers (I had to run some tests to check what they were!)

### DIFF
--- a/src/dvsim/testplan.py
+++ b/src/dvsim/testplan.py
@@ -16,14 +16,37 @@ from tabulate import tabulate
 
 
 class Result:
-    """The results for a single test."""
+    """The results for a job."""
 
-    def __init__(self, name, passing=0, total=0, job_runtime=None, simulated_time=None) -> None:
+    def __init__(
+        self,
+        name: str,
+        *,
+        passing: int = 0,
+        total: int = 0,
+        job_runtime_s: float | None = None,
+        simulated_time_us: float | None = None,
+    ) -> None:
+        """Construct a Result with the given parameters.
+
+        Args:
+          name: The name of the test.
+
+          passing: The number of runs that passed (possibly different seeds).
+
+          total: The number of runs that happened (will be at least as large as
+          passing).
+
+          job_runtime_s: If not None, the number of seconds taken by the job.
+
+          simulated_time_us: If not None, the simulated time in microseconds.
+
+        """
         self.name = name
         self.passing = passing
         self.total = total
-        self.job_runtime = job_runtime
-        self.simulated_time = simulated_time
+        self.job_runtime = job_runtime_s
+        self.simulated_time = simulated_time_us
         self.mapped = False
 
 
@@ -752,7 +775,7 @@ class Testplan:
         test_results = []
         for item in test_results_:
             try:
-                tr = Result(item["name"], item["passing"], item["total"])
+                tr = Result(item["name"], passing=item["passing"], total=item["total"])
                 test_results.append(tr)
             except KeyError:
                 sys.exit(1)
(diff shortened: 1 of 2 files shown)


src/dvsim/testplan.py
```diff
@@ -16,9 +16,31 @@
...
 
-    def __init__(self, name, passing=0, total=0, job_runtime=None, simulated_time=None) -> None:
+    def __init__(
+        self,
+        name: str,
+        passing: int = 0,
+        total: int = 0,
+        job_runtime: float | None = None,
+        simulated_time: float | None = None,
+    ) -> None:
```
AlexJones0, line 29: One suggestion I might have would be to make most of the arguments here keyword-only:


```suggestion
    def __init__(
        self,
        name: str,
        *
        passing: int = 0,
        total: int = 0,
        job_runtime: float | None = None,
        simulated_time: float | None = None,
    ) -> None:
```

This is because currently you can make a `Result` like:
```py
Result("example test", 3, 5, 19.20, 13.41)
```
which isn't particularly descriptive to read, so you need to search through the definition. This change would instead mean you need to write:
```py
Result("example test", passing=3, total=5, job_runtime=19.20, simulated_time=13.41)
```

I'll leave it up to you whether you prefer that or not though.

src/dvsim/testplan.py
```diff
@@ -16,9 +16,31 @@
...
+          name: The name of the test.
+
+          passing: The number of runs that passed (possibly different seeds).
+
+          total: The number of runs that happened (will be at least as large as
+          passing).
+
+          job_runtime: If not None, the number of seconds taken by the job.
+
+          simulated_time: If not None, the simulated time in microseconds.
```
AlexJones0: It doesn't need to be changed in this PR (as I appreciate it would be a widespread change across the code base) - but IMO these parameters should be suffixed by their units (e.g. `job_runtime_s` and `simulated_time_us`) to avoid the confusion entirely.

src/dvsim/testplan.py
```diff
@@ -16,9 +16,31 @@
...
+    ) -> None:
+        """Construct a Result with the given parameters.
+
+        Args:
+          name: The name of the test.
+
+          passing: The number of runs that passed (possibly different seeds).
+
+          total: The number of runs that happened (will be at least as large as
+          passing).
```
AlexJones0, line 38: Another nice idea might be to add setters for `passing` and `total` to this class to enforce this invariant and error otherwise - but I appreciate that's probably outside of the scope of the improvements in this PR, which is just doing some cleanup.

AlexJones0 (approved):
Thanks @rswarbrick!

src/dvsim/testplan.py
rswarbrick: Good idea. I didn't actually know that syntax (so just had [some reading](https://peps.python.org/pep-3102/)): it's rather neat!

src/dvsim/testplan.py
rswarbrick: Good point. Since I'm touching these lines, let's make the change here, at least.

rswarbrick: (I'm changing the argument names. The fields might have to wait)

src/dvsim/testplan.py
rswarbrick: Yeah, maybe: it might not be the cleanest solution though: we only actually pass those values in one place, I think (and it's to a constructor).

It might be that a better long-term solution is to switch to a dataclass.

PR #217, opened 2026-05-21
docs: Improve documentation comment for testplan.Element.has_tags

rswarbrick:
This was partly motivated by the fact I couldn't immediately see whether the test was an AND or an OR.

### DIFF
--- a/src/dvsim/testplan.py
+++ b/src/dvsim/testplan.py
@@ -101,18 +101,19 @@ class Element:
             msg = f"'tags' key in {self} is not a list."
             raise ValueError(msg)
 
-    def has_tags(self, tags: set) -> bool:
-        """Checks if the provided tags match the tags originally set.
+    def has_tags(self, tags: set[str]) -> bool:
+        """Return true if the element matches the provided tags.
 
-        tags is a list of tags that are we are filtering this testpoints with.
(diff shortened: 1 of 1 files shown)


src/dvsim/testplan.py
```diff
@@ -83,18 +83,19 @@ def _validate(self) -> None:
...
-    def has_tags(self, tags: set) -> bool:
-        """Checks if the provided tags match the tags originally set.
+    def has_tags(self, tags: set[str]) -> bool:
+        """Return true if the element matches the provided tags.
 
-        tags is a list of tags that are we are filtering this testpoints with.
-        Tags may be preceded with `-` to exclude the testpoints that contain
-        that tag.
+        The element should match every item in the set of tags. If one of these
+        elements is preceded with "-", the meaning is negated (so the element
```
AlexJones0: I think this might be intended to be:
```suggestion
        tags is preceded with "-", the meaning is negated (so the element
```
(or "`items`" also makes sense in the given context)

AlexJones0 (approved):
LGTM modulo one nit.

src/dvsim/testplan.py
rswarbrick: Oops: thanks!

PR #218, opened 2026-05-21
refactor: Remove outdated (and unused) default repo_top argument

rswarbrick:
This code was based on dvsim living inside the OpenTitan repository. Fortunately, it is never used: we always pass a third argument to _parse_testplan.

Also, fail more understandably if there is an hjson parse error or an invalid filename passed to _parse_hjson.

Finally, simplify the Testplan constructor so that it always expects a name for its testplan (rather than exiting the program silently if there isn't one).

We also change that argument to explicitly have type str (as opposed to Path) and explain why: the string can have suffix values that filter the set of tests.

### DIFF
--- a/src/dvsim/testplan.py
+++ b/src/dvsim/testplan.py
@@ -65,7 +65,7 @@ class Element:
     def __init__(self, raw_dict) -> None:
         """Initialize the testplan element.
 
-        raw_dict is the dictionary parsed from the HJSon file.
+        raw_dict is the dictionary parsed from the Hjson file.
         """
         # 'tags' is an optional field in addition to the mandatory self.fields.
         self.tags = []
@@ -261,7 +261,7 @@ class Testplan:
 
     @staticmethod
     def _parse_hjson(filename):
-        """Parses an input file with HJson and returns a dict."""
+        """Parses an input file with Hjson and returns a dict."""
         try:
             return hjson.load(Path(filename).open())
         except OSError:
@@ -275,7 +275,7 @@ class Testplan:
         """Create testplan elements from the list of raw dicts.
 
         kind is either 'testpoint' or 'covergroup'.
-        raw_dicts_list is a list of dictionaries extracted from the HJson file.
+        raw_dicts_list is a list of dictionaries extracted from the Hjson file.
         """
         items = []
         item_names = set()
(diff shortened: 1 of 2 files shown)


src/dvsim/testplan.py
```diff
@@ -244,15 +243,9 @@ class Testplan:
...
-        try:
-            return hjson.load(Path(filename).open())
-        except OSError:
-            pass
-        except hjson.scanner.HjsonDecodeError:
-            pass
-        sys.exit(1)
+    def _parse_hjson(filename: Path) -> dict:
+        """Parse an hjson file at the given path and return it as a dict."""
+        return hjson.load(Path(filename).open())
```
AlexJones0: This was a problem in the original code, but we should probably fix it while we're here.

Reference counting probably means that this issue will not be a problem, but we should be explicitly closing the file handle after opening it rather than leaving it open.

A better approach would be:

```suggestion
        """Parse an hjson file at the given path and return it as a dict."""
        with Path(filename).open("r", encoding="utf-8") as f:
            return hjson.load(f)
```

Or maybe even better as:

```suggestion
        """Parse an hjson file at the given path and return it as a dict."""
        return hjson.loads(Path(filename).read_text(encoding="utf-8"))
```
(to let Pathlib handle the read file mode for us).

src/dvsim/testplan.py
```diff
@@ -295,64 +288,37 @@ def _get_percentage(value, total) -> str:
...
 
-        filename is the HJson file that captures the testplan. It may be
-        suffixed with tags separated with a colon delimiter to filter the
-        testpoints. For example: path/too/foo_testplan.hjson:bar:baz
-        repo_top is an optional argument indicating the path to top level repo
-        / project directory. It is used with filename arg.
-        name is an optional argument indicating the name of the testplan / DUT.
-        It overrides the name set in the testplan HJson.
+        Args:
+          filename: Describes the HJson file that captures the testplan. This
```
AlexJones0:
```suggestion
          filename: Describes the Hjson file that captures the testplan. This
```
(a couple more times in these changed docs - I appreciate it was in the original but it'd be nice to be consistent).

src/dvsim/testplan.py
```diff
@@ -295,64 +288,37 @@ def _get_percentage(value, total) -> str:
...
+                    is a string, rather than a Path object, because it may be
+                    suffixed with tags separated with a colon delimiter to
+                    filter the testpoints.
```
AlexJones0: The explanation is great, but given that we later do
```py
        filename = Path(split[0])
        tags = set(split[1:])
```
I'd suggest the name isn't great anyway and we should rename it. Perhaps something like:
* `tagged_filename`
* `filename_with_tags`
* `file_spec` / `testplan_file_spec`

etc. etc. - take your pick.

AlexJones0 (approved):

src/dvsim/testplan.py
rswarbrick: Thanks, that makes sense (and the scope of the file object hadn't occurred to me). I guess the former solution might be nice if we consumed the file as a stream with some sort of reducing operation. But that's not what's going on here, so the second is definitely better (*another* thing that I understood for the first time this morning...)

src/dvsim/testplan.py
rswarbrick: I agree. I've added a staging commit that fixes this.

src/dvsim/testplan.py
rswarbrick: Good point! I'll go for the first.

rswarbrick:
Damn! Wrong project's commit naming convention! Take two...

rswarbrick:
And now for plurals. Take three...

PR #219, opened 2026-05-21
refactor: Collect tps/cgs with tighter types in testplan.py

rswarbrick:
There's no functional change, but this avoids needing to pass around constructors. As a result, the inferred types for Testplan.testpoints and Testpaln.covergroups are tighter, meaning that some following code is now well-typed.


AlexJones0 (approved):
Nice! This looks like a great improvement. Thanks @rswarbrick!

PR #221, opened 2026-05-21
refactor: Tidy up how wildcards get expanded in testpoints

rswarbrick:
I think the existing implementation had a bit of a problem if you had multiple wildcard patterns that all expanded to lists of values. Presumably no-one ever cared, but it also wasn't very easy to see how the code worked.

### DIFF
--- a/src/dvsim/testplan.py
+++ b/src/dvsim/testplan.py
@@ -189,31 +189,89 @@ class Testpoint(Element):
             msg = f"'tests' key in {self} is not a list."
             raise ValueError(msg)
 
-    def do_substitutions(self, substitutions) -> None:
+    @staticmethod
+    def _expand_str(pattern: str, new_vals: Sequence[str], txt: str) -> list[str]:
+        """Return a list of copies of txt with each expansion of pattern.
+
+        Note that if txt has multiple copies of the pattern then they are
+        replaced "diagonally". If pattern is "x", new_vals is ["0", "1"] and
+        txt is "x x" then the result is ["0 0", "1 1"].
+
+        Args:
+          pattern: A string for which we should search.
+
+          new_vals: A list of values to use as replacements.
+
+          txt: The string in which to replace the given pattern.
+
+        """
+        return [txt.replace(pattern, v) for v in new_vals]
+
+    @staticmethod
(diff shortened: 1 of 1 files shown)


src/dvsim/testplan.py
```diff
@@ -173,31 +174,87 @@ def _validate(self) -> None:
...
+        from the substitutions argument.
+
+        If a value in the substitutions argument is a list, the substitution
+        will make multiple testpoints from a templated value. For example, if
+        substitutions is {'a': ['x', 'y']} and self.tests is ['test-{a}'] then
+        self.tests will become ['test-x', 'test-y'].
+
+        A wildcard use with no match in substitutions will be replaced by an
+        empty string (in a single item). For example, if substitutions is {}
+        and self.tests is ['test-{a}'] then self.tests will become ['test-'].
```
AlexJones0, line 245: Aside: the original implementation feels unintuitive - I feel like most similar systems would retain one copy of the unsubstituted text (i.e. just leave it as `['test-{a}']`).

src/dvsim/testplan.py
```diff
@@ -173,31 +174,87 @@ def _validate(self) -> None:
...
+
+          txt_list: List of strings to expand
+
+        """
+        ret: list[str] = []
+        for txt in txt_list:
+            ret.extend(Testpoint._expand_str(pattern, new_vals, txt))
+        return ret
+
+    def do_substitutions(self, substitutions: dict, reserved_names: Sequence[str]) -> None:
```
AlexJones0: Can we make this `substitutions: dict[str, str | list[str]]` or is this too strong of a typing to enforce here at this stage? If not, at least `substitutions: dict[str, Any]`?

AlexJones0 (approved):

src/dvsim/testplan.py
rswarbrick: I'm not completely sure, but this might be an intentional feature. It also guarantees that test names don't have stray curly braces in them at the end.

I think any improvement to this would probably need to be separate work.

src/dvsim/testplan.py
rswarbrick: Because the dictionary fundamentally comes from a parsed hjson file, I think we'll need the weaker type. But that's probably better than what I had.

PR #222, opened 2026-05-21
docs: Improve documentation of Testplan.get_percentage

rswarbrick:
Also, add two error conditions (for fractions that don't make sense).

### DIFF
--- a/src/dvsim/testplan.py
+++ b/src/dvsim/testplan.py
@@ -336,12 +336,19 @@ class Testplan:
             known_names.add(elem.name)
 
     @staticmethod
-    def _get_percentage(value, total) -> str:
-        """Returns a string representing percentage upto 2 decimal places."""
-        if total == 0:
-            return "-- %"
-        perc = value / total * 100 * 1.0
-        return f"{round(perc, 2):.2f} %"
+    def _get_percentage(value: int, total: int) -> str:
+        """Format a fraction (value / total) as a float with up to 2 decimal places.
+
+        Both arguments should be non-negative and value should be at most equal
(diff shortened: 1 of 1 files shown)


src/dvsim/testplan.py
```diff
@@ -288,12 +288,19 @@ def _create_testplan_elements(kind: str, raw_dicts_list: list, tags: set) -> lis
...
+        Both arguments should be non-negative and value should be at most equal
+        to total. If total is zero, this is reported as "-- %".
+
+        """
+        if value > total:
+            raise ValueError("Cannot represent fraction over 100%")
+        if value < 0:
+            raise ValueError("Cannot represent negative fraction")
+
+        return "-- %" if total == 0 else f"{round(100 * value / total, 2):.2f} %"
```
AlexJones0: I'm pretty sure the `round` here is pointless - that's what the `2` in the format specifier is doing.
Python actually has a nifty format specifier for almost exactly this case:

```suggestion
        return "--%" if total == 0 else f"{value / total:.2%}"
```
But then you drop the space between the number and the % symbol. You could do:

```suggestion
        return "--%" if total == 0 else f"{value / total:.2%}".replace("%", " %")
```
Though that's about as much effort as the original.

AlexJones0 (approved):

src/dvsim/testplan.py
rswarbrick: Good point about the rounding! Let's keep the multiplication by 100 (for now, at least). It's possibly more explicit anyway: I'm a bit ambivalent about using format specifiers that change the number.

PR #223, opened 2026-05-24
Fix status shown for combined instrumentation jobs

AlexJones0:
In the HTML instrumentation report, the graph showing the longest tests (where each tests is many different jobs) had an issue where, for combined job bars (rendered to reduce the number of bars being stored/shown on lower render profiles for performance reasons), the status of the first/shortest job would be reported in the metadata of the tooltip shown on hovering over the bar.

This was instead intended to not show a status at all, since there might be variation in the outcomes of these combined jobs. Fix this by adding the omitted argument.


machshev (approved):
Thanks @AlexJones0

PR #229, opened 2026-06-04
fix: Allow testplans to have entries that are booleans

rswarbrick:
It turns out that there is even a schema for (sival) testplans in OpenTitan at hw/lint/sival_testplan_schema.hjson. After this change, that schema is stricter than the requirements of dvsim.


machshev (approved):
Thanks @rswarbrick!

PR #230, opened 2026-06-04
fix: Improve error message when failing to parse a testplan

rswarbrick:
This way, the error message at the bottom of the console log is rather more helpful. In the example I just ran:

RuntimeError: Error when parsing testplan at /home/rjs/work/opentitan/hw/top_earlgrey/data/chip_testplan.hjson: The otp_mutate field in the testplan element with name manuf_cp_unlock_raw is neither a string nor a list of strings.


machshev (approved):
Thanks @rswarbrick

PR #231, opened 2026-06-04
fix: Fix broken tp/cg filtering/update in testplan.py

rswarbrick:
I accidentally removed the filtering with f328cbcc. Then (worse!) removed the update to the class variable in 78c0fab.

This commit puts both back, and also now distinguishes between "there are no testpoints" and "you have filtered out all of the testpoints".


machshev (approved):
Thanks @rswarbrick!

PR #232, opened 2026-06-04
refactor: Improve and document Testplan.get_stage_regressions

rswarbrick:
(*I was inspired by doing some simple bug fixes, so I picked the next undocumented function in the file...*)

This was prompted by trying to understand the function in order to document what it did. Changes to the code:

 - Sort the tests associated with a stage instead of using the default ordering that was presumably based on the way the set of tests got built up from testpoints.

 - Explicitly use Testpoint.stages for the class variable instead of tp.stages: I wasted five minutes trying to figure out what was going on here.

 - Avoid the "ms" name (the thing that is now called a development stage was called a milestone in OpenTitan until 2022: see d1aeac6).

 - Use setdefault instead of a defaultdict

 - Iterate over items rather than iterating over keys and doing a lookup each time.

### DIFF
--- a/src/dvsim/testplan.py
+++ b/src/dvsim/testplan.py
@@ -6,7 +6,6 @@
 
 import re
 import sys
-from collections import defaultdict
 from collections.abc import Sequence
 from pathlib import Path
 from typing import Any, TextIO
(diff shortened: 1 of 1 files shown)


machshev (approved):
Thanks @rswarbrick!

PR #234, opened 2026-06-21
Fix slurm.py to work correctly with dvsim

sha-ron:
Fixed `unexpected end of file` syntax error.
A few more issues that were fixed:
•	dvsim hangs indefinitely after job completes
•	No pass/fail summary displayed
•	Timer/progress display frozen

### DIFF
--- a/src/dvsim/launcher/slurm.py
+++ b/src/dvsim/launcher/slurm.py
@@ -34,6 +34,7 @@ class SlurmLauncher(Launcher):
 
         # Popen object when launching the job.
         self.process = None
+        self._log_file = None
         self.slurm_log_file = f"{self.job_spec.log_path}.slurm"
 
     def _do_launch(self) -> None:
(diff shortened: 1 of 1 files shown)


src/dvsim/launcher/slurm.py
```diff
@@ -57,34 +58,34 @@ def _do_launch(self) -> None:
...
         try:
-            with pathlib.Path(self.slurm_log_file).open("w") as out_file:
-                out_file.write(f"[Executing]:\n{self.job_spec.cmd}\n\n")
-                out_file.flush()
-
-                log.info(f"Executing slurm command: {slurm_cmd}")
-                self.process = subprocess.Popen(
-                    shlex.split(slurm_cmd),
-                    bufsize=4096,
-                    universal_newlines=True,
```
AlexJones0, line 75: Perhaps I'm missing something - do we not still need `universal_newlines=True` (or `text=True`) since we're trying to write to a log file opened in `w` mode and not `wb`?

AlexJones0 (commented):
Thanks! I can't test this locally at the moment but the changes seem like reasonable fixes to me. Just one comment that I'm not entirely sure about and then I think this is good to go in.

src/dvsim/launcher/slurm.py
sha-ron: Thank you for the review @AlexJones0.
Slurm works after applying the changes in this PR (at least for me). So I guess anyway it is better then what we have now that fails on compilation. 
As for the note: honestly, AI claimed it is not needed (see explanation below) and indeed it worked fine without it. But I added it again just in case.
"When Popen redirects stdout/stderr to a file object, it uses dup2 on the underlying file descriptor — the child process writes directly to the fd, bypassing the Python file object entirely. So text vs binary mode on the Python side doesn't matter".

AlexJones0: That makes sense, thanks. I think you are right that in this case it doesn't matter, albeit the `subprocess` docs aren't the most clear about this detail :sweat_smile:

AlexJones0 (approved):

sha-ron:
@AlexJones0 @rswarbrick can you please tag a new dvsim version and use it in https://github.com/lowRISC/opentitan/blob/master/python-requirements.txt#L43 ?

rswarbrick:
@sha-ron: You can! :-) Every version of dvsim gets a "release", so the only work required is to bump the version of dvsim required by the OpenTitan repository.

PR #237, opened 2026-07-07
feat: Adds CDC results parsing for Meridian in the CDC flow

nsw-lowrisc:
This commit adds parsing of the Meridian CDC reports. The parser is using the meridiancdc.py file in the tools directory, so the parser in the local repo is not required, however, if the report_cmd variable is set in the hjson this will be overridden and the parser called by make.

NOTE: the buckets have changed, adding waived and tofix, so the corresponding common_cdc_cfg.hjson file should be updated in line. This required a fix to msg_buckets.py to avoid an exception.

### DIFF
--- a/src/dvsim/flow/cdc.py
+++ b/src/dvsim/flow/cdc.py
@@ -2,17 +2,45 @@
 # Licensed under the Apache License, Version 2.0, see LICENSE for details.
 # SPDX-License-Identifier: Apache-2.0
 
-"""Class describing lint configuration object."""
+"""Class describing CDC configuration object."""
+
+import logging
+from collections.abc import Sequence
 
 from dvsim.flow.lint import LintCfg
+from dvsim.job.data import CompletedJobStatus
+from dvsim.tool.meridiancdc import cdc_parse_main
+from dvsim.utils import subst_wildcards
+
+# Get the global logger definition
+log = logging.getLogger(__name__)
 
 
 class CdcCfg(LintCfg):
-    """Derivative class for linting purposes."""
+    """Derivative class from LintCfg for parsing CDC logs."""
 
     flow = "cdc"
 
+    # Override the __init__ for a CDC run
+    # only names at the moment
     def __init__(self, flow_cfg_file, hjson_data, args, mk_config) -> None:
         super().__init__(flow_cfg_file, hjson_data, args, mk_config)
 
         self.results_title = f"{self.name.upper()} CDC Results"
+
+    # Override the _gen_results_for_cfg to generate the results.hsjon file
+    # As _gen_results_for_cfg expects the results.hjson to be there
+    # The results parsing is inserted before calling LintCfg _gen_results_for_cfg.
+    def _gen_results_for_cfg(self, results: Sequence[CompletedJobStatus]) -> None:
+        # Call the log parser for each build mode if it is not part of the build
+        # Note report_cmd can be added to the meridiancdc.hjson file if a local
+        # version of the parser is preferred and available.
+        if self.report_cmd is None or len(self.report_cmd) == 0:
+            for mode in self.build_modes:
+                result_path = subst_wildcards(self.build_dir, {"build_mode": mode.name})
+                result_file = f"{result_path}/results.hjson"
+                cdc_parse_main(["--repdir", result_path, "--outfile", result_file])
+        else:
+            log.info(f"For results.hjson Using external parser: {self.report_cmd}")
+
+        return super()._gen_results_for_cfg(results)
--- a/src/dvsim/msg_buckets.py
+++ b/src/dvsim/msg_buckets.py
@@ -193,4 +193,6 @@ class MsgBuckets:
             if not isinstance(signatures, list):
                 msg = f"Signatures in {k} must be a list of strings"
                 raise RuntimeError(msg)
-            self.buckets[k].signatures.extend(signatures)
+            # check the key is in the bucket list to avoid an exception
+            if k in self.buckets:
+                self.buckets[k].signatures.extend(signatures)
--- a/src/dvsim/tool/meridiancdc.py
+++ b/src/dvsim/tool/meridiancdc.py
@@ -1,0 +1,285 @@
+#!/usr/bin/env python3
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Parses Meridian CDC report and dumps filtered messages in hjson format."""
+
+import argparse
+import logging
+import re
+import sys
+from pathlib import Path
+
+from dvsim.linting.parser import LintParser
+
+# For stand alone
+# Check if this file is the main entry point
+IS_STANDALONE = __name__ == "__main__"
+if IS_STANDALONE:
+    # Get the absolute path of the directory one level up (the project root)
+    project_root = str(Path(__file__).resolve().parent.parent.parent)
+    # Add it to Python's search path if it isn't already there
+    if project_root not in sys.path:
+        sys.path.insert(0, project_root)
+
+
+# Get the global logger definition
+log = logging.getLogger(__name__)
+
+
+def extract_rule_patterns(file_path: Path):
+    """This parses the CDC summary table to get the message totals,
+    rule names and corresponding severities.
+    """
+    rule_patterns = []
+    full_file = ""
+    try:
+        with Path(file_path).open() as f:
+            full_file = f.read()
+    except OSError:
+        # We will attempt read this file again in a second pass to parse out
+        # the details, this error will get caught and reported.
+        pass
+
+    category = ""
+    severity = ""
+    known_rule_names = {}
+    # total_msgs = 0
+    # extract the summary table
+    m = re.findall(
+        r"^Summary of Policy: NEW((?:.|\n|\r\n)*)Rule Details of Policy: NEW",
+        full_file,
+        flags=re.MULTILINE,
+    )
+    if m:
+        # step through the table and identify rule names and their
+        # category and severity
+        for line in m[0].split("\n"):
+            if re.match(r"^POLICY\s+NEW", line):
+                continue
+                # total = re.findall(r"^POLICY\s+NEW\s+([0-9]+)", line)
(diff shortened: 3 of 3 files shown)


src/dvsim/flow/cdc.py
```diff
@@ -2,17 +2,45 @@
 # Licensed under the Apache License, Version 2.0, see LICENSE for details.
 # SPDX-License-Identifier: Apache-2.0
 
-"""Class describing lint configuration object."""
+"""Class describing CDC configuration object."""
+
+import logging
```
machshev, line 7: We have a configured logger available as `from dvsim.logging import log`. This does some extra setup to add new log levels. Could we use that please?

src/dvsim/tool/meridiancdc.py
```diff
@@ -0,0 +1,285 @@
+#!/usr/bin/env python3
+# Copyright lowRISC contributors (OpenTitan project).
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Parses Meridian CDC report and dumps filtered messages in hjson format."""
+
+import argparse
+import logging
```
machshev, line 9: Same here

src/dvsim/tool/meridiancdc.py
```diff
@@ -0,0 +1,285 @@
...
+
+# For stand alone
+# Check if this file is the main entry point
+IS_STANDALONE = __name__ == "__main__"
+if IS_STANDALONE:
+    # Get the absolute path of the directory one level up (the project root)
+    project_root = str(Path(__file__).resolve().parent.parent.parent)
+    # Add it to Python's search path if it isn't already there
+    if project_root not in sys.path:
+        sys.path.insert(0, project_root)
```
machshev, line 24: Why are we doing this? This looks really hacky... what are you trying to achieve?

src/dvsim/flow/cdc.py
```diff
@@ -2,17 +2,45 @@
...
+from collections.abc import Sequence
 
 from dvsim.flow.lint import LintCfg
+from dvsim.job.data import CompletedJobStatus
+from dvsim.tool.meridiancdc import cdc_parse_main
+from dvsim.utils import subst_wildcards
+
+# Get the global logger definition
+log = logging.getLogger(__name__)
```
machshev, line 16:
```suggestion
from dvsim.logging import log
```

src/dvsim/tool/meridiancdc.py
```diff
@@ -0,0 +1,285 @@
...
+
+
+# Get the global logger definition
+log = logging.getLogger(__name__)
```
machshev, line 28:
```suggestion
from dvsim.logging import log
```

src/dvsim/tool/meridiancdc.py
```diff
@@ -0,0 +1,285 @@
...
+
+def extract_rule_patterns(file_path: Path):
+    """This parses the CDC summary table to get the message totals,
+    rule names and corresponding severities.
+    """
+    rule_patterns = []
+    full_file = ""
+    try:
+        with Path(file_path).open() as f:
+            full_file = f.read()
```
machshev, line 39:
```suggestion
        file_path.read_text()
```

src/dvsim/tool/meridiancdc.py
```diff
@@ -0,0 +1,285 @@
...
+    severity = ""
+    known_rule_names = {}
+    # total_msgs = 0
+    # extract the summary table
+    m = re.findall(
+        r"^Summary of Policy: NEW((?:.|\n|\r\n)*)Rule Details of Policy: NEW",
+        full_file,
+        flags=re.MULTILINE,
+    )
+    if m:
```
machshev, line 55: Generally I'd go for the early return which means you can de-dent the rest of the function

```
if not m:
  return []
```

src/dvsim/tool/meridiancdc.py
```diff
@@ -0,0 +1,285 @@
...
+        with Path(file_path).open() as f:
+            full_file = f.read()
+    except OSError:
+        # We will attempt read this file again in a second pass to parse out
+        # the details, this error will get caught and reported.
+        pass
+
+    category = ""
+    severity = ""
+    known_rule_names = {}
```
machshev, line 47: Maybe move these down the function closer to where they are used?

src/dvsim/tool/meridiancdc.py
```diff
@@ -0,0 +1,285 @@
...
+    # Add it to Python's search path if it isn't already there
+    if project_root not in sys.path:
+        sys.path.insert(0, project_root)
+
+
+# Get the global logger definition
+log = logging.getLogger(__name__)
+
+
+def extract_rule_patterns(file_path: Path):
```
machshev, line 31: Could you add typing for the return type... it looks like `List[Tuple[str,str]]`?
With this added then `pyright` is a lot more useful.

src/dvsim/tool/meridiancdc.py
```diff
@@ -0,0 +1,285 @@
...
+                else:
+                    msg_group = category + "_" + severity
+                    known_rule_names.update({name: msg_group})
+                    rule_patterns.append((msg_group, rf"^{name}:\s+\d+.*"))
+
+    return rule_patterns
+
+
+# Reuse the lint parser, but add more buckets.
+class CdcParser(LintParser):
```
machshev, line 108: Class docstring?

src/dvsim/tool/meridiancdc.py
```diff
@@ -0,0 +1,285 @@
...
+if __name__ == "__main__":
+    # Configure Logging log to console if running stand alone
+    # logging format layout
+    log_layout = "[%(levelname).1s %(asctime)s %(filename)s:%(lineno)d] %(message)s"
+    time_layout = "%y%m%d %H:%M:%S"
+    console_handler = logging.StreamHandler()
+    console_handler.setFormatter(logging.Formatter(f"{log_layout}", datefmt=time_layout))
+    active_handlers = [console_handler]
+
+    logging.basicConfig(level=logging.INFO, handlers=active_handlers)
```
machshev, line 283: This should be covered if you use the dvsim logger?

src/dvsim/msg_buckets.py
```diff
@@ -193,4 +193,6 @@ def load_hjson(self, result_path: Path) -> None:
             if not isinstance(signatures, list):
                 msg = f"Signatures in {k} must be a list of strings"
                 raise RuntimeError(msg)
-            self.buckets[k].signatures.extend(signatures)
+            # check the key is in the bucket list to avoid an exception
```
machshev, line 196: Do we want to log a warning if it's not? Is this an error case?

machshev (commented):
Thanks @nsw-lowrisc!
Some nits but the general concept seems fine.

src/dvsim/tool/meridiancdc.py
nsw-lowrisc: Thanks @machshev for the review.

I think it is best to resolve this point first, as others depend on it. 

The other parsers in this directory (lint/RDC/...) are imported from the "in-tree" version of dvsim. The in-tree versions are still run stand alone from the make file. This, the CDC parser, was initial derived from the standalone RDC parser. This is the only one that can now be called either way, from dvsim or the make file. This where the IS_STANDALONE logic comes from and also other things like the logging (I added this in the standalone version, so should be corrected) and parsing the input parameters with argparse. 

So, if it is not desirable to keep the possibility of running stand alone then all the extra stuff should be removed. I am OK with either, what ever fits best into the bigger picture.

src/dvsim/msg_buckets.py
nsw-lowrisc: This only to prevent, dvsim crashing out completely with an esoteric error when there a mismatch between the buckets in the hjson and python. The idea is only to smooth transition when adopting the new parser. I would say that long term the way the buckets are managed should be revised, that is only one definition needed. 

I can add a warning massage.

src/dvsim/tool/meridiancdc.py
machshev: Yes makes sense for the version in OT repo it's something that needs fixing across all of the parsers. We just haven't got around to fixing this up. However this code can never be called except as standalone, because it's installed to python's site packages?

So perhaps clean this version up?

src/dvsim/flow/cdc.py
nsw-lowrisc: Updated now

src/dvsim/tool/meridiancdc.py
nsw-lowrisc: done

src/dvsim/tool/meridiancdc.py
nsw-lowrisc: I have refactored to remove the standalone possibility

src/dvsim/flow/cdc.py
nsw-lowrisc: done

src/dvsim/tool/meridiancdc.py
nsw-lowrisc: done

src/dvsim/tool/meridiancdc.py
nsw-lowrisc: done

src/dvsim/tool/meridiancdc.py
nsw-lowrisc: gone now using the logging config

src/dvsim/msg_buckets.py
nsw-lowrisc: Added a warning if they buckets dont match

nsw-lowrisc:
I have made the suggested changes, hopefully thats every thing fixed

src/dvsim/tool/meridiancdc.py
nsw-lowrisc: updated now

src/dvsim/tool/meridiancdc.py
nsw-lowrisc: done

src/dvsim/tool/meridiancdc.py
nsw-lowrisc: done

PR #241, opened 2026-08-06
Long line deadlock fix

### LINKED ISSUE #240: LocalRuntimeBackend deadlocks when a job emits a line ≥ 64 KiB on stdout/stderr
## Summary

dvsim hangs indefinitely (0% CPU, no progress, no timeout) when a launched tool writes a single line of ≥ 64 KiB without a newline to stdout or stderr.
The job, and therefore the whole dvsim invocation, deadlocks.
The root cause is in the async local runtime backend's pipe-reader (`src/dvsim/runtime/local.py`), which reads tool output line-by-line through an asyncio.StreamReader whose default line limit is 64 KiB.

## Affected component

- src/dvsim/runtime/local.py → LocalRuntimeBackend._log_from_pipe()
- Observed on dvsim 1.34.1 and 1.50.0 (the relevant code is identical in both).

## How to reproduce

Run any flow whose tool emits a very long single line (no \n) on stdout.
It reproduces reliably with a Xcelium/JasperGold UNR run:

```
dvsim hw/ip/hmac/dv/hmac_sim_cfg.hjson --tool xcelium --cov-unr
```

JasperGold's UNR report contained a 116,398-character line, which is what triggered it in practice.
A minimal repro is any subprocess that does `print("x" * 100_000)`.

## Symptoms / observable state

- dvsim sits forever in its asyncio event loop; the CLI status stays "in progress".
- The launched tool (e.g. xrun) is blocked in write() on its stdout pipe (anon_pipe_write).
- The tool's log file stops growing partway through a large output burst.
- In dvsim's epoll set, the stdout pipe fd is absent while the stderr pipe fd is still registered — the tell-tale sign that the stdout transport was paused and never resumed.

## Root cause

_log_from_pipe() consumes the subprocess stream with line iteration:

```
async for line in stream:                       # StreamReader.readline() under the hood
    decoded = line.decode("utf-8", errors="surrogateescape")
    handle.log_file.write(decoded)
    handle.log_file.flush()
```

The subprocess is created with:

```
proc = await asyncio.create_subprocess_exec(
    *shlex.split(job.cmd),
    stdout=asyncio.subprocess.PIPE,
    stderr=asyncio.subprocess.PIPE,     # note: no limit= → default 64 KiB StreamReader
    env=env,
)
```

Because no limit= is passed, each StreamReader uses the default _DEFAULT_LIMIT = 2**16 (64 KiB).
When a single line exceeds that limit, StreamReader.readline() raises ValueError (from an internal LimitOverrunError).
_log_from_pipe() only catches asyncio.CancelledError, so the exception propagates and the reader task dies.

The deadlock then cascades:

1. Stdout reader task dies → nobody drains the stdout StreamReader.
2. Its buffer stays above the high-water mark, so asyncio calls
transport.pause_reading(), which removes the stdout fd from the event loop.
3. The 64 KiB OS pipe fills; the child process blocks in write() and never
exits.
4. _monitor_job() is await-ing process.wait(). With no timeout_secs configured
for the flow (e.g. cov_unr), it waits forever.
5. The child stops reading its own downstream tools, so the entire tool
process tree wedges behind the full pipe.

The unretrieved reader-task exception is silent (surfaced at most as an
asyncio "Task exception was never retrieved" warning), so there is no error
message pointing at the cause.

## Proposed solution

Read the streams in fixed-size chunks rather than lines.
StreamReader.read(n) returns up to n bytes and has no line-length limit, eliminating the failure mode entirely.
An incremental UTF-8 decoder is used so multibyte characters split across a chunk boundary are decoded correctly (previously guaranteed because each read ended on a newline).

```
async def _log_from_pipe(self, handle, stream):
    """Write piped asyncio subprocess stream contents to a job's log file."""
    if stream is None or not handle.log_file:
        return
    decoder = codecs.getincrementaldecoder("utf-8")(errors="surrogateescape")
    try:
        while True:
            chunk = await stream.read(65536)
            if not chunk:
                break
            handle.log_file.write(decoder.decode(chunk))
            handle.log_file.flush()
        tail = decoder.decode(b"", final=True)
        if tail:
            handle.log_file.write(tail)
            handle.log_file.flush()
    except asyncio.CancelledError:
        pass
```

(Requires import codecs.)

## Optional hardening (defence-in-depth, recommended alongside the fix)

- Broaden the except in _log_from_pipe() so any future reader-task failure logs an error and returns instead of silently deadlocking the job.
- Plumb a job timeout_secs through the flows that currently pass None (e.g. cov_unr), so any residual hang self-recovers instead of blocking indefinitely.

## Notes

- The legacy synchronous launcher (launcher/nc.py) redirected tool stdout/stderr straight to the log file descriptor, so it was structurally immune to this. The bug was introduced with the async runtime rewrite.
- The fix is a pure logging/transport change: byte-for-byte log output is unchanged for normal (newline-terminated) tool output.

hcallahan-lowrisc:
## Description

Fixes #240
Commit messages describe the fixes, and 5 new unit tests are added in support of this.

## Checklist

### DIFF
--- a/src/dvsim/runtime/local.py
+++ b/src/dvsim/runtime/local.py
@@ -5,6 +5,7 @@
 """Legacy launcher adapter interface for the new async scheduler design."""
 
 import asyncio
+import codecs
 import contextlib
 import os
 import pty
@@ -51,6 +52,15 @@ class LocalRuntimeBackend(RuntimeBackend):
 
     INTERACTIVE_TEE_READ_SIZE = 1024  # Read 1024 bytes to balance efficiency with responsiveness
 
+    # StreamReader buffer limit for subprocess pipes, matching the asyncio default of 64 KiB.
+    # Only relevant to line-oriented reads (readline/readuntil)
+    SUBPROCESS_STREAM_LIMIT = 64 * 2**10  # 64 KiB
+
+    # Size of each fixed-size read from a subprocess pipe in `_log_from_pipe`. Independent of the
+    # StreamReader limit above: `StreamReader.read(n)` returns up to n bytes and never raises on a
+    # long line, so this only trades off syscall frequency against per-read memory.
+    SUBPROCESS_READ_CHUNK_SIZE = 64 * 2**10  # 64 KiB
+
     def __init__(
         self,
         *,
@@ -86,13 +96,29 @@ class LocalRuntimeBackend(RuntimeBackend):
         """Write piped asyncio subprocess stream contents to a job's log file."""
         if stream is None or not handle.log_file:
             return
+        # Read fixed-size chunks - iterating lines can cause unexpected errors if the
+        # subprocess emits very long lines which overflows the StreamReader buffer.
+        # Uses an incremental decoder to handle multibyte characters that straddle
+        # a chunk boundary.
+        decoder = codecs.getincrementaldecoder("utf-8")(errors="surrogateescape")
         try:
-            async for line in stream:
-                decoded = line.decode("utf-8", errors="surrogateescape")
-                handle.log_file.write(decoded)
+            while True:
+                chunk = await stream.read(self.SUBPROCESS_READ_CHUNK_SIZE)
+                if not chunk:
+                    break
+                handle.log_file.write(decoder.decode(chunk))
+                handle.log_file.flush()
+            tail = decoder.decode(b"", final=True)
+            if tail:
+                handle.log_file.write(tail)
                 handle.log_file.flush()
         except asyncio.CancelledError:
             pass
+        except Exception:  # noqa: BLE001
+            log.exception(
+                "Error while streaming subprocess output to log file for job '%s'.",
+                handle.spec.full_name,
+            )
 
     async def _monitor_job(self, handle: LocalJobHandle) -> None:
         """Wait for subprocess completion and emit a completion event."""
@@ -238,6 +264,7 @@ class LocalRuntimeBackend(RuntimeBackend):
                     # useful to make this behaviour optional on some global `IoPolicy`.
                     stdout=asyncio.subprocess.PIPE,
                     stderr=asyncio.subprocess.PIPE,
+                    limit=self.SUBPROCESS_STREAM_LIMIT,
                     env=env,
                 )
             except BlockingIOError:
--- a/tests/test_runtime.py
+++ b/tests/test_runtime.py
@@ -4,20 +4,30 @@
 
 """Test the DVSim Runtime Backends."""
 
+import asyncio
 import importlib
+import io
+import shlex
+import sys
 from collections.abc import Generator
+from pathlib import Path
+from types import SimpleNamespace
 
 import pytest
 from hamcrest import (
     assert_that,
     calling,
+    contains_string,
     equal_to,
     instance_of,
     raises,
 )
 from pytest_mock import MockerFixture
 
+from dvsim.job.status import JobStatus
+from dvsim.runtime import local as local_module
 from dvsim.runtime.backend import RuntimeBackend
+from dvsim.runtime.data import JobCompletionEvent
 from dvsim.runtime.legacy import LegacyLauncherAdapter
 from dvsim.runtime.local import LocalRuntimeBackend
 from dvsim.runtime.registry import (
@@ -26,7 +36,7 @@ from dvsim.runtime.registry import (
     register_backend,
     register_legacy_launcher_backend,
 )
-from tests.test_scheduler import MockLauncher
+from tests.test_scheduler import MockLauncher, job_spec_factory
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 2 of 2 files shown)


src/dvsim/runtime/local.py
```diff
@@ -86,10 +87,26 @@ async def _log_from_pipe(
...
             return
+        # Read fixed-size chunks rather than iterating lines. `async for line in
+        # stream` uses StreamReader.readline(), which raises LimitOverrunError
+        # (surfaced as ValueError) once a single line exceeds the StreamReader
+        # limit (64 KiB by default). A tool emitting a very long line without a
+        # newline (e.g. Jasper's UNR report) would otherwise crash this reader
+        # task; with no consumer the transport is paused, the OS pipe fills, and
+        # the subprocess deadlocks writing to it. StreamReader.read() has no such
+        # limit. An incremental decoder handles multibyte characters that straddle
+        # a chunk boundary.
```
AlexJones0: Nit: I think the commit history is fine for seeing what came before, no need to reproduce the history & error context in this comment. I think it's useful to have, but I'd probably shorten it to the useful context. 

Something like:
```
# Read fixed-size chunks - iterating lines can cause unexpected errors if the
# subprocess emits very long lines which overflows the StreamReader buffer.
# Uses an incremental decoder to handle multibyte characters that straddle
# a chunk boundary.
```

WDYT?

src/dvsim/runtime/local.py
```diff
@@ -86,10 +87,26 @@ async def _log_from_pipe(
...
+        decoder = codecs.getincrementaldecoder("utf-8")(errors="surrogateescape")
         try:
-            async for line in stream:
-                decoded = line.decode("utf-8", errors="surrogateescape")
-                handle.log_file.write(decoded)
+            while True:
+                chunk = await stream.read(65536)
```
AlexJones0: Nit: We should try to avoid the inline constant - maybe this should be derived from `SUBPROCESS_STREAM_LIMIT`?

src/dvsim/runtime/local.py
```diff
@@ -52,6 +52,12 @@ class LocalRuntimeBackend(RuntimeBackend):
 
     INTERACTIVE_TEE_READ_SIZE = 1024  # Read 1024 bytes to balance efficiency with responsiveness
 
+    # StreamReader buffer limit for subprocess pipes. asyncio defaults to 64 KiB, at which point
+    # readline() raises LimitOverrunError on an over-long line. `_log_from_pipe` reads fixed-size
+    # chunks and no longer relies on this, but we raise the limit as defense-in-depth in case any
+    # line-oriented reading is reintroduced, and to keep the buffer generous for bursty tools.
```
AlexJones0: I see the argument for defense-in-depth but I'm not sure that this is particularly useful for the memory cost - this seems like it might be a *little* excessively large - keep in mind that we might be running hundreds of parallel jobs, and that we have a reader for both `stderr` and `stdout`.

Otherwise, the same nit again on the comment - lets avoid talking about the usage of this constant in `_log_from_pipe` - something like:
```
# Set a larger (16 MiB) limit for subprocess StreamReader buffers to defend against
# potential reads of large lines without chunking.
```

src/dvsim/runtime/local.py
```diff
@@ -110,6 +116,16 @@ async def _log_from_pipe(
...
+        except Exception:  # noqa: BLE001
+            # Defense-in-depth: never let an unexpected error kill this reader
+            # task silently. A dead reader stops draining the pipe, so asyncio
+            # pauses the transport, the OS pipe fills, and the subprocess
+            # deadlocks writing to it (see comment above). Log it so the failure
+            # surfaces rather than manifesting as a mysterious hang.
+            log.exception(
+                "Error while streaming subprocess output to log file for job '%s'.",
+                handle.spec.full_name,
+            )
```
AlexJones0, line 121: Would recommend logging the exception here as well (`except Exception as e`, etc.)

But actually, I think a better fix (untested), would be to change the [`asyncio.gather(...)`](https://github.com/lowRISC/dvsim/blob/836ff253276c78424b76507b7d9e00e6bd08311b/src/dvsim/runtime/local.py#L130) to use [`return_exceptions=False`](https://docs.python.org/3/library/asyncio-task.html#running-tasks-concurrently), to stop the exceptions being aggregated.

AlexJones0 (commented):
Thanks - mostly looks good to me, I just have a couple of small suggestions.

src/dvsim/runtime/local.py
hcallahan-lowrisc: Yes I agree, I should have cleaned this up before. WILCO

src/dvsim/runtime/local.py
hcallahan-lowrisc: Agreed.

src/dvsim/runtime/local.py
hcallahan-lowrisc: `log.exception` prints the trace as well as the message. This is now checked as part of a new test as well:
```
$ python -m pytest "tests/test_runtime.py::TestLocalBackendStreaming::test_reader_error_is_logged_not_raised" -s -o log_cli=true --log-cli-level=ERROR

<...>

tests/test_runtime.py::TestLocalBackendStreaming::test_reader_error_is_logged_not_raised [E 260807 12:06:43 plugin:178] Error while streaming subprocess output to log file for job 'job'.
Traceback (most recent call last):
  File "/home/harry/projects/dvsim/src/dvsim/runtime/local.py", line 109, in _log_from_pipe
    handle.log_file.write(decoder.decode(chunk))
    ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/harry/projects/dvsim/tests/test_runtime.py", line 230, in write
    raise OSError("simulated disk-full")
OSError: simulated disk-full
PASSED

```

src/dvsim/runtime/local.py
hcallahan-lowrisc: Created a new constant for this and set it to 64KiB (the default) as well.

hcallahan-lowrisc:
Addressed review comments and added 5 unit tests covering the failure mode / new implementation.

PR #242, opened 2026-08-17
feat: report EDA tool version from a dedicated version query

hcallahan-lowrisc:
## Description

Replace the hardcoded "unknown" tool version in reports with a value probed from the tool at runtime. Each SimTool plugin declares a VersionQuery (command + regex) as data, and a single generic runner executes it, so adding a tool needs no new logic. Any failure falls back to "unknown".

A tool version is now printed inside the final report badge. E.g.
image

## Checklist

### DIFF
--- a/src/dvsim/sim/flow.py
+++ b/src/dvsim/sim/flow.py
@@ -43,7 +43,7 @@ from dvsim.sim.report import gen_reports
 from dvsim.sim_results import BucketedFailures, SimResults
 from dvsim.test import Test
 from dvsim.testplan import Testplan
-from dvsim.tool.utils import get_sim_tool_plugin
+from dvsim.tool.utils import get_sim_tool_plugin, query_tool_version
 from dvsim.utils import TS_FORMAT, rm_path
 from dvsim.utils.fs import relative_to
 from dvsim.utils.git import git_https_url_with_commit
@@ -716,7 +716,7 @@ class SimCfg(FlowCfg):
             url=url,
             revision_info=self.revision,
         )
-        tool = ToolMeta(name=self.tool.lower(), version="unknown")
+        tool = ToolMeta(name=self.tool.lower(), version=query_tool_version(self.tool) or "unknown")
 
         build_seed = self.build_seed if not self.run_only else None
 
--- a/src/dvsim/sim/tool/base.py
+++ b/src/dvsim/sim/tool/base.py
@@ -5,8 +5,9 @@
 """EDA simulation tool interface."""
 
 from collections.abc import Mapping, Sequence
+from dataclasses import dataclass
 from pathlib import Path
-from typing import TYPE_CHECKING, Protocol, runtime_checkable
+from typing import TYPE_CHECKING, ClassVar, Protocol, runtime_checkable
 
 from dvsim.job.data import JobSpec
 from dvsim.sim.data import CoverageMetrics
@@ -14,12 +15,31 @@ from dvsim.sim.data import CoverageMetrics
 if TYPE_CHECKING:
     from dvsim.job.deploy import Deploy
 
-__all__ = ("SimTool",)
+__all__ = ("SimTool", "VersionQuery")
+
+
+@dataclass(frozen=True)
+class VersionQuery:
+    """Declarative description of how to query an EDA tool for its version.
+
+    Attributes:
+        cmd: the command that makes the tool print its version.
+        pattern: a regex applied (in multiline mode) to the combined
+            stdout/stderr of ``cmd``. The first capture group is used as the
+            version string.
+
+    """
+
+    cmd: str
+    pattern: str
 
 
 @runtime_checkable
 class SimTool(Protocol):
     """Simulation tool interface required by the Sim workflow."""
+
+    version_query: ClassVar[VersionQuery | None]
+    """How to query this tool's version, or ``None`` if unsupported."""
 
     @staticmethod
     def get_cov_summary_table(cov_report_path: Path) -> tuple[Sequence[Sequence[str]], str]:
--- a/src/dvsim/sim/tool/vcs.py
+++ b/src/dvsim/sim/tool/vcs.py
@@ -7,10 +7,11 @@
 import re
 from collections.abc import Mapping, Sequence
 from pathlib import Path
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, ClassVar
 
 from dvsim.job.data import JobSpec
 from dvsim.sim.data import CodeCoverageMetrics, CoverageMetrics
+from dvsim.sim.tool.base import VersionQuery
 
 if TYPE_CHECKING:
     from dvsim.job.deploy import Deploy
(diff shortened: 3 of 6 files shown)


src/dvsim/sim/tool/base.py
```diff
@@ -5,22 +5,42 @@
...
+
+@dataclass(frozen=True)
+class VersionQuery:
+    """Declarative description of how to query an EDA tool for its version.
+
+    Attributes:
+        cmd: the command that makes the tool print its version.
+        pattern: a regex applied (in multiline mode) to the combined
+            stdout/stderr of ``cmd``. The first capture group is used as the
+            version string.
```
rswarbrick, line 29: I'm not a Python expert, but I'm slightly surprised by this. Is there a reason not to use this text to decorate the fields, rather than as a string at the top?

src/dvsim/sim/tool/vcs.py
```diff
@@ -21,6 +22,12 @@
 class VCS:
     """Implement VCS tool support."""
 
+    # `vcs -id` reports a line like: "Compiler version = VCS X-2025.06-SP2-1_Full64".
```
rswarbrick: I'd probably either explain the 64 bit thing, or give "vcs -full64 -id" in the string here.

src/dvsim/sim/tool/vcs.py
```diff
@@ -21,6 +22,12 @@
...
+    version_query: ClassVar[VersionQuery | None] = VersionQuery(
+        cmd="vcs -full64 -id",
+        pattern=r"^Compiler version\s*=\s*(?:VCS\s+)?(\S+)",
```
rswarbrick: I take it that some versions don't include "VCS"? Maybe this needs a comment explaining why the extra group is here.

Alternatively, would something like this work?
```
^Compiler version\s*=.*\s([^ ]+)$
```

src/dvsim/sim/tool/xcelium.py
```diff
@@ -22,6 +23,12 @@
 class Xcelium:
     """Implement Xcelium tool support."""
 
+    # `xrun -version` reports a line like: "TOOL:   xrun(64)        24.03-s007".
+    version_query: ClassVar[VersionQuery | None] = VersionQuery(
+        cmd="xrun -version",
+        pattern=r"^TOOL:\s*xrun\(\d+\)\s+(\S+)",
```
rswarbrick: As with the VCS example, I'd suggest being floppier. How about something like this?
```
^TOOL:.*xrun.*\s([^ ]+)$
```

src/dvsim/sim/flow.py
```diff
@@ -716,7 +716,7 @@ def _gen_json_results(
             url=url,
             revision_info=self.revision,
         )
-        tool = ToolMeta(name=self.tool.lower(), version="unknown")
+        tool = ToolMeta(name=self.tool.lower(), version=query_tool_version(self.tool) or "unknown")
```
rswarbrick, line 719: This should probably be split over 2 lines to avoid falsy strings with "is None".

src/dvsim/tool/utils.py
```diff
@@ -4,20 +4,30 @@
...
-__all__ = ("get_sim_tool_plugin",)
+__all__ = ("get_sim_tool_plugin", "query_tool_version")
 
 _SUPPORTED_SIM_TOOLS = {
     "vcs": VCS,
     "xcelium": Xcelium,
     "z01x": Z01X,
 }
 
+# EDA tools should respond to a `--version`-style query near-instantly, but guard
```
rswarbrick: So that it doesn't sound like "EDA tools should do X but might guard Y", I'd suggest splitting the string. "... query near-instantly. This timeout is just to ensure that a hung/misconfigured tool can't block things forever."

src/dvsim/tool/utils.py
```diff
@@ -31,3 +41,73 @@ def get_sim_tool_plugin(tool: str) -> SimTool:
...
+    tool is not on PATH, a non-EDA host, a timeout, ...) is treated as "version
+    unknown" rather than an error: querying the version must never break a run.
+
+    Results are cached for the lifetime of the process since a tool's version is
+    invariant across a single dvsim invocation; this dedupes the query when
+    multiple configs share a tool.
+    """
+    try:
+        result = subprocess.run(  # noqa: S603
+            shlex.split(cmd),
```
rswarbrick, line 63: This looks odd to me. Since `subprocess.run` will do the splitting, there's no reason to do it ourselves - things are already globbed together, so we don't know any more than the library code.

src/dvsim/tool/utils.py
```diff
@@ -31,3 +41,73 @@ def get_sim_tool_plugin(tool: str) -> SimTool:
...
+            capture_output=True,
+            text=True,
+            timeout=_VERSION_QUERY_TIMEOUT_S,
+            check=False,
+        )
+    except (OSError, subprocess.SubprocessError) as e:
+        log.debug("Failed to query tool version via '%s': %s", cmd, e)
+        return None
+
+    return result.stdout + result.stderr
```
rswarbrick, line 73: Nit: If the order doesn't matter, I'd suggest doing them the other way round (because the result is more like what normally happens on the terminal)

src/dvsim/tool/utils.py
```diff
@@ -31,3 +41,73 @@ def get_sim_tool_plugin(tool: str) -> SimTool:
...
+    output = run(query.cmd)
+    if output is None:
+        return None
+
+    match = re.search(query.pattern, output, re.MULTILINE)
+    if match is None:
+        log.debug("Could not parse %s version from output of '%s'", tool, query.cmd)
+        return None
+
+    return match.group(1).strip()
```
rswarbrick: I'm not sure, but does it make sense to catch `IndexError`s here and spit out an explicit error to make it easier to write the EDA backend? Or is this just "the author should read the docs..."? :-)

rswarbrick (commented):
This looks really sensible to me. Lots of nitty comments (sorry), but it looks good.

src/dvsim/sim/tool/base.py
hcallahan-lowrisc: Only because that's the google style, which I think we're trying to follow here. https://google.github.io/styleguide/pyguide.html#384-classes

src/dvsim/sim/tool/vcs.py
hcallahan-lowrisc: Made the regex floppier.

src/dvsim/sim/tool/xcelium.py
hcallahan-lowrisc: Ditto above.

src/dvsim/sim/flow.py
hcallahan-lowrisc: I updated the implementation to return None and log if the regex match is `""`. Hence decided to keep the use of `or` here.

src/dvsim/tool/utils.py
hcallahan-lowrisc: Updated to two sentences. It may just be me, but my brain refuses to parse the guarding clause as being associated with the EDA tool. The `, but` is just too strong for me :0

src/dvsim/tool/utils.py
hcallahan-lowrisc: subprocess.run only splits a string argument when `shell=True`, so we do need the shlex here. I think this is pretty idiomatic.

src/dvsim/tool/utils.py
hcallahan-lowrisc: Sorry I don't follow this argument, I'm pretty sure each stream just races each other to write into the terminal? Either way, it's inconsequential here.

src/dvsim/tool/utils.py
hcallahan-lowrisc: Good idea, we now catch it, log, and return None in this case.

PR #243, opened 2026-08-18
feat: score the vPlan from test results as well as coverage

martin-velay:
Linked to https://github.com/lowRISC/dvplan/pull/117

## What this does

The `cov_vplan` job already annotated a vPlan from the coverage database. It now also
annotates from the regression's test results, so a nightly says how much of the
verification plan is met and not only how much of the design was covered. A test the
plan asked for and the regression never ran shows as a hole rather than going missing.

dvsim writes those results to `dv_evidence.json` and passes it to dvplan alongside the
coverage report. It also passes through a path to hand-written inspection records if the
cfg names one.

Nothing runs unless the sim cfg sets `vplan`.

## Three decisions worth reviewing

**The evidence comes from a new scheduler callback, not from the end-of-run results.**
Everything it records is already in `run_results`, but that only exists once the scheduler
has finished, and `cov_vplan` is a scheduled job that needs the file before then. The
alternative is to stop scheduling it and run it at `gen_results` time, which would drop
the callback, the collector and the `FlowCfg` hook. I kept it scheduled so the step keeps
its row in the status table and can fail like any other job.

**`needs_all_dependencies_passing` became `DependencyPolicy`.** The bool had two states
and this job needs a third. `ALL_PASSING` is the default and `ANY_PASSING` is what
`CovMerge` always did, so neither changes. `CovVPlan` takes `ALWAYS`, because a regression
where nothing passed is exactly the case the plan has to describe, and under either
existing policy it was killed instead of scored: with `--cov` it has one dependency, so
anything that stopped the coverage report also stopped the plan.

**dvsim defines the evidence format, not dvplan.** `doc/dv_evidence.md` specifies it and
the pydantic models are normative. dvsim is what produces the file and is the public repo,
so a consumer can be written against a spec rather than against whichever tool was built
first. dvplan is one consumer.

## Dependencies and degradation

Needs the linked dvplan PR first: it defines the `dv_evidence` source name and the
multi-source behaviour. Whether dvplan is installed is decided by the job's own script on
the machine it runs on, so a checkout without it warns and passes.

## The credential fix rides along

`git_https_url_with_commit` returns the `origin` remote verbatim today, so a CI checkout
with `https://oauth2:<token>@host/org/repo` already publishes the token in the JSON and
HTML reports. This PR adds another archived artefact carrying the same URL, so the fix is
first in the series. It reverts cleanly on its own.

## Behaviour change

A cfg that sets `vplan` and runs without `--cov` previously got no vPlan job. It now gets
one, scored from test results alone.

## Verified

`ruff format --check` and `ruff check --config=ruff-ci.toml` clean, `pytest --strict` 396
passed, `license_check.py` clean, `pyright` 182 errors unchanged, every commit tested
standalone. Run end to end against a real hmac regression in an OpenTitan tree.

### DIFF
--- a/src/dvsim/flow/base.py
+++ b/src/dvsim/flow/base.py
@@ -22,6 +22,7 @@ from dvsim.fusesoc import rewrite_fusesoc_opts
 from dvsim.job.data import CompletedJobStatus, JobSpec, WorkspaceConfig
 from dvsim.job.status import JobStatus
 from dvsim.logging import log
+from dvsim.scheduler.core import OnJobCompletionCb
 from dvsim.scheduler.resources import UnknownResourcePolicy
 from dvsim.scheduler.runner import (
     build_default_scheduler_backend,
@@ -527,8 +528,17 @@ class FlowCfg(ABC):
                 interactive=self.interactive,
                 backend=backend,
                 resource_manager=resource_manager,
+                on_job_completed=self.job_completion_callback(),
             )
         )
+
+    def job_completion_callback(self) -> OnJobCompletionCb | None:
+        """Return an observer for jobs reaching a terminal state, or None to observe nothing.
+
+        Asked for once, as the scheduler is built. Observing is opt-in and the base flow declines,
+        so nothing here knows what any one flow does with the outcomes.
+        """
+        return None
 
     @abstractmethod
     def gen_results(self, results: Sequence[CompletedJobStatus]) -> None:
--- a/src/dvsim/job/data.py
+++ b/src/dvsim/job/data.py
@@ -10,6 +10,7 @@ capture the results of the job run.
 """
 
 from collections.abc import Callable, Mapping, Sequence
+from enum import Enum
 from pathlib import Path
 from typing import TypeAlias
 
@@ -20,10 +21,28 @@ from dvsim.report.data import IPMeta, ToolMeta
 
 __all__ = (
     "CompletedJobStatus",
+    "DependencyPolicy",
     "JobSpec",
     "JobStatusInfo",
     "WorkspaceConfig",
 )
+
+
+class DependencyPolicy(Enum):
+    """When a job may start, given how the jobs it depends on ended."""
+
+    ALL_PASSING = "all_passing"
+    """Start only if every dependency passed, which is right for a job consuming their output."""
+
+    ANY_PASSING = "any_passing"
+    """Start if at least one dependency passed, for a job gathering whatever results exist."""
+
+    ALWAYS = "always"
+    """Start once every dependency is terminal, whatever they concluded.
+
+    For a job whose input is the outcome itself rather than an artefact a dependency produced, so
+    a regression where nothing passed is still the thing it has to report on.
+    """
 
 
 class WorkspaceConfig(BaseModel):
@@ -92,8 +111,8 @@ class JobSpec(BaseModel):
 
     dependencies: list[str]
     """Full names of the other Jobs that this one depends on."""
-    needs_all_dependencies_passing: bool
-    """Wait for dependent jobs to pass before scheduling."""
+    dependency_policy: DependencyPolicy
+    """What the jobs this one depends on must have concluded before it may be scheduled."""
     weight: int
     """Weight to apply to the scheduling priority."""
     timeout_mins: float | None
--- a/src/dvsim/job/deploy.py
+++ b/src/dvsim/job/deploy.py
@@ -17,6 +17,13 @@ from dvsim.job.status import JobStatus
 from dvsim.job.time import JobTime
 from dvsim.logging import log
 from dvsim.report.data import IPMeta, ToolMeta
+from dvsim.report.dv_evidence import write_evidence
+from dvsim.report.vplan import (
+    VPLAN_DIR,
+    VPlanInputs,
+    overall_coverage,
+    shell_command,
+)
 from dvsim.test import Test
 from dvsim.tool.utils import get_sim_tool_plugin
 from dvsim.utils import (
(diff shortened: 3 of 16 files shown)


src/dvsim/utils/git.py
```diff
@@ -68,7 +93,7 @@ def git_origin_url(path: Path | None = None) -> str | None:
     if "origin" not in [remote.name for remote in r.remotes]:
         return None
 
-    return r.remote("origin").url
+    return strip_url_credentials(r.remote("origin").url)
```
machshev: Good spot... generally these tokens are short lived tokens and only required for internal runs on private repos. But still this isn't great.

Thanks for fixing!

src/dvsim/flow/base.py
```diff
@@ -491,9 +491,20 @@ def deploy_objects(self) -> Sequence[CompletedJobStatus]:
...
 
+    def on_job_completed(
+        self, spec: JobSpec, status: JobStatus, reason: JobStatusInfo | None
+    ) -> None:
+        """Observe a job reaching a terminal state. Flows that care override this.
+
+        Called by the scheduler as each job finishes, so a flow can record an outcome while the
+        run is still going. Not abstract, since observing this is opt-in.
+        """
+        del spec, status, reason
```
machshev: I think I'd prefer a `# noqa: ARG002` rather than a noop `del` just to satisfy the linter. Alternatively just don't create the callback in this commit, create it when it's used. The scheduler arg is optional defaulting to `None`, so you don't need to create or pass it in this commit.

src/dvsim/job/deploy.py
```diff
@@ -379,6 +386,11 @@ def is_equivalent_job(self, item: "Deploy") -> bool:
         log.verbose('Deploy job "%s" is equivalent to "%s"', item.name, self.name)
         return True
 
+    @property
+    def log_path(self) -> Path:
+        """Path to the log this job writes."""
+        return Path(f"{self.odir}/{self.target}.log")
+
```
machshev: Personally I prefer the previous inline version unless the new property is going to be used elsewhere, and then it creates a good single source of truth. However even if that is the case `Deploy` objects are deprecated and hopefully soon replaced by the `JobSpec` objects. Then the creating of the JobSpec can be moved upstream to the `Flow` objects with a factory method style interface or something similar.

So instead of getting the `log_path` from the `Deploy` object it should really be got from the created `JobSpec` object.

src/dvsim/report/vplan.py
```diff
@@ -0,0 +1,172 @@
...
+def shell_command(inputs: VPlanInputs) -> str:
+    """Build the bash command that prepares and annotates the vPlan.
+
+    Returned as one `bash -c` string because a scheduled job runs a shell command. `set -e` and the
+    `&&` mean a broken annotation shows as a failed job rather than a silently missing score.
+    """
+    if shutil.which("dvplan") is None:
+        # Warn and pass, so a checkout without dvplan does not fail every regression naming a vPlan
+        warning = "WARNING: dvplan is not installed on PATH. Skipping vPlan annotation."
+        return f"/usr/bin/env bash -c {shlex.quote(f'echo {shlex.quote(warning)}')}"
```
machshev: This looks like it's using bash call to print warnings? Python should be able to do that direct.

src/dvsim/report/vplan.py
```diff
@@ -0,0 +1,172 @@
...
+
+    # The vPlan sits at <ip_root>/<something>/<vplan>, so its grandparent is the IP root that
+    # `prepare_vplan` traces specifications against.
+    ip_root = inputs.vplan.parent.parent
```
machshev, line 120: Why are we building a shell script here? Python can call these commands individually and do better error handling around them if it does.

src/dvsim/sim/flow.py
```diff
@@ -158,10 +159,15 @@ def __init__(self, flow_cfg_file, hjson_data, args, mk_config) -> None:
         self.cov_report_dir = ""
         self.cov_report_page = ""
 
-        # Options for vPlan processing
+        # Options for vPlan processing. Extracted from the hjson cfg, and declared here so a type
+        # checker knows they exist. Nothing happens unless `vplan` names a plan
+        self.vplan: str = ""
```
machshev, line 166: Why are we extending the flow base class here? Does DVPlan run with all flow types? If it's Sim flow only then this should be in the Sim flow class not the base class. The Deploy subclass for VPlan https://github.com/lowRISC/dvsim/pull/243/changes/7f2bd4d0e3734f077adcb3bfcc6cb6bd01a7282c#diff-f5ec728727bbae78e93d3e9155c5a57d18dad7fa7ceb893fff9886dab16999a0R1062 has `SimCfg` typing... but this is now being added to the flow base class?

machshev: Admittedly this is an existing confusion, it seems DVSim was originally designed for Sim flow only and the others are an after thought bolt on as far as the design is concerned. But we should try and scope additions to the right level.

src/dvsim/sim/flow.py
```diff
@@ -684,6 +697,45 @@ def gen_results(self, results: Sequence[CompletedJobStatus]) -> None:
...
+    def on_job_completed(
+        self, spec: JobSpec, status: JobStatus, reason: JobStatusInfo | None
+    ) -> None:
+        """Record each run's outcome as the scheduler concludes it.
+
+        Fed from the scheduler rather than from a job callback, because only the scheduler sees a job
+        it cancelled before dispatching it.
+        """
+        self.run_evidence.record(spec, status, reason)
+
```
machshev: This does mean the post processing is being interleaved with the scheduled jobs. Other steps in the flows use either:

1. `JobSpec` submission to the scheduler to perform dependency based scheduling
2. `gen_results` report post processing based on the `CompletedJobStatus` of each job

This adds a new mechanism that is interleaved with (1) but not visible to the scheduler or the metrics we record in the scheduler. I'm tempted to suggest that this is refactored into one of the two existing mechanisms.

So if the work in this callback is reporting, then maybe add it to the `gen_results` and filter out the `VPlan` jobs from the `CompletedJobStatus` collection and do the post processing there. If there is significant work involved i.e. calling external processes to do parsing and such like then I'd suggest it should be moved into it's own `Job` and pass through the scheduler as a substantial piece of work.

Message passing between jobs should ideally be done via the filesystem in the scratch directory. That means we can later implement restarting of partial runs, or rerunning of individual steps. When the original runtime memory for the previous steps have been lost, the scratch directory working files are still there.

src/dvsim/job/data.py
```diff
@@ -20,12 +21,30 @@
...
+    """Start if at least one dependency passed, for a job gathering whatever results exist."""
+
+    ALWAYS = "always"
+    """Start once every dependency is terminal, whatever they concluded.
+
+    For a job whose input is the outcome itself rather than an artefact a dependency produced, so
+    a regression where nothing passed is still the thing it has to report on.
+    """
+
+
```
machshev, line 47: Looks like a nice improvement! Thanks...

machshev (commented):
Thanks @martin-velay, great work in general. Some parts of the architecture I'm not so sure about, given the ongoing refactoring work. But hopefully we can resolve these fairly quickly.

src/dvsim/flow/base.py
martin-velay: You're right, fixed

src/dvsim/job/deploy.py
martin-velay: You're right, property reverted to inline

src/dvsim/sim/flow.py
martin-velay: You're right, should be fixed now

src/dvsim/report/vplan.py
machshev: Ah, this is in the remote runner. So this makes sense that it's preparing commands to run remotely.
We could have a remote dvsim instance and that might make sense going forward for some jobs but it's a bigger change than the scope of this PR.

src/dvsim/report/vplan.py
machshev: Answered above... for the remote runner.

src/dvsim/report/vplan.py
```diff
@@ -0,0 +1,172 @@
...
+# Scratch subdirectory the annotated plan and its report are written to. Unchanged, so an existing
+# link to the report still resolves
+VPLAN_DIR = "cov_vplan"
+
+ANNOTATED_HJSON = "vplan_annotated.hjson"
+ANNOTATED_HTML = "vplan_annotated.html"
+EVIDENCE_JSON = "dv_evidence.json"
+
+
+@dataclass(frozen=True)
```
machshev: It would be nice to use pydantic model instead of dataclass here. Should be a drop-in replacement almost.

martin-velay: Should be done now

PR #256, opened 2026-08-31
feat: expose a build's options to the run step as {build_opts_file}

martin-velay:
## Description

A run step that compiles for itself instead of loading the build's snapshot has to compile the way the build did.  Every build now records its options in build_opts.f, named by the hjson as {build_opts_file}.


src/dvsim/job/deploy.py
```diff
@@ -36,6 +36,10 @@
     "Deploy",
 )
 
+# Name of the file that records the options a build used. It sits in the build directory, and
+# SimCfg turns it into the {build_opts_file} substitution variable the HJson can name.
+BUILD_OPTS_FILE_NAME = "build_opts.f"
```
rswarbrick: If we're using wildcards, this probably needs to be possible to *supply* from the hjson file as well. I don't think there's currently the notion of a read-only variable.

src/dvsim/job/deploy.py
```diff
@@ -509,6 +515,23 @@ def _set_attrs(self) -> None:
         if self.sim_cfg.args.build_timeout_mins is not None:
             self.build_timeout_mins = self.sim_cfg.args.build_timeout_mins
 
+    def _write_build_opts_file(self) -> None:
+        """Record the options this build used, in the build directory.
+
+        A run step that compiles and elaborates for itself, rather than loading the snapshot the
+        build produced, has to compile the way the build did: build modes and individual cfgs each
+        add their own defines, include paths and libraries, and a run compiled with a different set
+        would simulate a differently configured design. The merged option list exists only here, so
```
rswarbrick: This sentence took me several goes to understand. Could you split it up?

src/dvsim/job/deploy.py
```diff
@@ -509,6 +515,23 @@ def _set_attrs(self) -> None:
...
+        it is written out in a form the tools accept as an option file, for the run step to read
+        back with {build_opts_file}.
```
rswarbrick: How does that wildcard value become visible from hjson code? I don't think it currently does?

src/dvsim/job/deploy.py
martin-velay: Is it better now?

src/dvsim/job/deploy.py
martin-velay: Good point, it should be fixed now

src/dvsim/job/deploy.py
martin-velay: It was not clear to me, so I give you Claude's answer:

> It falls out of the normal substitution pass: FlowCfg._expand() calls find_and_substitute_wildcards(self.__dict__, self.__dict__, ...), so every cfg attribute is a wildcard, and SimCfg._expand() sets build_opts_file before calling super()._expand(). Substitution is recursive, so an hjson run_opts: ["-f {build_opts_file}"] resolves through the nested {build_dir} to the full path.

src/dvsim/job/deploy.py
```diff
@@ -509,6 +511,25 @@ def _set_attrs(self) -> None:
...
+        The merged option list exists only here, so write it out as a file the tools accept in
+        place of command line options. The run step names that file with {build_opts_file}, which
+        SimCfg sets and the HJson may override.
+        """
+        opts_file = Path(self.build_opts_file)
+        opts_file.parent.mkdir(parents=True, exist_ok=True)
+        opts_file.write_text(
+            "".join(f"{opt.strip()}\n" for opt in self.build_opts if opt.strip()),
+            encoding="UTF-8",
+        )
```
rswarbrick, line 531: I squinted at this for a few minutes. I'm was sad about the repeated calls to `opt.strip()` and the fact that we concatenate a bunch of lines, appending `\n` to each. It took me a while to spot that the reason to do so is to ensure there's a `\n` at the end of the file.

I think the following structure is probably a bit simpler to understand, and also splits up "render an option" from "write a bunch of options to a file". What do you think?
```python
        maybe_options = [opt.strip() for opt in self.build_opts]
        options = [opt for opt in maybe_options if opt]

        opts_file = Path(self.build_opts_file)
        opts_file.parent.mkdir(parents=True, exist_ok=True)
        opts_file.write_text("\n".join(options) + "\n", encoding="UTF-8")
```

src/dvsim/job/deploy.py
```diff
@@ -509,6 +511,25 @@ def _set_attrs(self) -> None:
         if self.sim_cfg.args.build_timeout_mins is not None:
             self.build_timeout_mins = self.sim_cfg.args.build_timeout_mins
 
+    def _write_build_opts_file(self) -> None:
+        """Record the options this build used, in the build directory.
```
rswarbrick, line 515: I spent a while trying to work out whether there was a more concise way to write this, and make a docstring that's easier to understand. I *think* this contains all the information we need to convey, and is much easier to understand. Does it look reasonable to you?

```python
        """Write build options to the build directory.

        Doing so allows a run step to recompile and elaborate for itself, without having to
        re-invoke dvsim from scratch. It would be complicated for an external tool to infer these
        options: they depend on lots of configuration files. Writing them out here solves that
        problem.

        """
```

src/dvsim/sim/flow.py
```diff
@@ -200,6 +201,13 @@ def _expand(self) -> None:
         if self.args.verbosity is not None:
             self.verbosity = self.args.verbosity
 
+        # Where each build records the options it compiled with, for a run step that compiles for
+        # itself rather than loading the snapshot the build produced, see
+        # CompileSim._write_build_opts_file(). The HJSON can name this path as {build_opts_file},
+        # and can also set it, to move the file or to share one between cfgs.
```
rswarbrick, line 207: Please could this comment move to the "declaration" of the field in constructor? (I know that dvsim isn't great at this! But here's an opportunity to do it right :-)

tests/job/test_deploy.py
```diff
@@ -202,6 +204,49 @@ def test_seed(
 
         assert_that(job.seed, equal_to(seed))
 
+    @staticmethod
+    def test_build_opts_file(tmp_path: Path) -> None:
+        """Test that a CompileSim records the options it built with, for the run step."""
+        build_dir = tmp_path / "build" / "dir"
+        job = _build_compile_sim(
+            sim_overrides={"build_dir": str(build_dir), "cov_db_dir": str(tmp_path / "cov")},
```
rswarbrick, line 212: What does `cov_db_dir` do here and in the next test?

rswarbrick (commented):
This looks nice, and I'm sorry for not getting back more quickly. I've got some suggestions about how to make the code structure a bit clearer, but I'm a fan of the change.

src/dvsim/job/deploy.py
rswarbrick: Thanks. I'd forgotten how the expansion works (which is embarrassing, because I think I wrote quite a bit of it!)

src/dvsim/job/deploy.py
martin-velay: Better indeed, thanks

src/dvsim/job/deploy.py
martin-velay: Done, thanks

src/dvsim/sim/flow.py
martin-velay: Moved

tests/job/test_deploy.py
martin-velay: Added a comment above

rswarbrick (approved):
Looks great to me: thanks.

martin-velay:
@machshev, is it OK for you too?

hcallahan-lowrisc (approved):
Thanks @martin-velay this LGTM

PR #257, opened 2026-09-02
fix: report the formal coverage columns the engine actually measured

martin-velay:
## Description

get_coverage read formal_coverage["formal"], ["stimuli"] and ["checker"], which are the columns a JasperGold run reports. A VC Formal run reports stimuli, coi and proof, so reporting one raised KeyError: 'formal' and took dvsim down with a traceback after the build job had already passed and written its results.

The columns are now taken from the report the tool's own parser wrote, and summary_header follows them, so the cross-cfg summary table names the columns the cfg's engine filled in rather than another engine's. Only stimuli is common to the two engines, so neither set can stand in for the other and there is nothing to harmonise onto.

This was reachable only once an OpenTitan-side bug was fixed. The vcformal parse-formal-report.py rejects the --exp-fail-path that common_formal_cfg.hjson always passes, so make failed, no results.hjson was written, result.get("coverage") returned None, and get_coverage took its "No coverage information found" branch instead.

tests/flow/ is new; nothing covered this module before.

AI-assisted (Claude Code) — reviewed and approved by author

### DIFF
--- a/src/dvsim/flow/formal.py
+++ b/src/dvsim/flow/formal.py
@@ -4,6 +4,7 @@
 
 from collections.abc import Sequence
 from pathlib import Path
+from typing import cast
 
 import hjson
 from tabulate import tabulate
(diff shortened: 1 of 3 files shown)


hcallahan-lowrisc (approved):
The changes make sense to me, thanks @martin-velay.
My only feedback is that the comment generation is a bit too wordy in my opinion, especially in the source code (tests is fine). The comments can be more compact, and don't need to give examples of usage and problems solved for instance. Those details are much better in the commit message or PR message I think.

PR #260, opened 2026-09-03
fix: strip credentials from the recorded git origin url

martin-velay:
CI checkouts leave the access token in the remote url, like https://oauth2:<token>@host/org/repo. We record that url in the run metadata and print it in the JSON and HTML reports, and those get archived, so the token leaks with them.

ssh urls keep their user. git@ is part of the address, so dropping it would give a url that no longer reaches the remote. Passwords are stripped whatever the scheme.

AI-assisted (Claude Code) - reviewed and approved by author

### DIFF
--- a/src/dvsim/utils/git.py
+++ b/src/dvsim/utils/git.py
@@ -5,12 +5,33 @@
 """Git utility functions."""
 
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


machshev (approved):
Thanks @martin-velay!

